PR #213, opened 2019-07-01
added documentation page with:  release name, generated date, license

cminter:
I wanted to know what version of the document I'm reading when I've printed out the document. So, in the PDF, I added a simple page before table of contents that contains:

- release version (uses TAG from Makefile)
- document generation date (uses default formatting)
- MIT license text (from durexforth.asm) with generated end date

(all the LaTeX commands used are from what I learned today, so please let me know if I used poor form)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,8 @@
 C1541   = c1541
 AS = acme
 TAG = `git describe --tags --abbrev=0 || svnversion --no-newline`
-TAG_DEPLOY_DOT = `git describe --tags --abbrev=0`
-TAG_DEPLOY = `git describe --tags --abbrev=0 | tr . _`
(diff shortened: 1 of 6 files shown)


cminter:
I realized now tag created date would be better than generation date

jkotlinski:
Excellent! OK, I'll wait for another commit then. Thanks!

cminter:
Last commit (df466a908081ea30f024904e339e1908faea81d6) uses the date from the current commit in the LaTeX doc.  Also, just puts a "-M" in the deploy filenames if you deploy having modified (uncommitted) files.

It doesn't modify the disk name.  I noticed there isn't any room left in the disk title.  I think that would even be a problem if you happened to have a normal release that has more than 3 digit characters (e.g. 1.6.10)

jkotlinski:
Thanks!

PR #227, opened 2020-01-19
Added * searching to v

burnsauce:
Beginner forth code, might not be the most efficient.

Pressing * over "bar" will match the next bar, but not "foobar" or "barf" which should be the expected behaviour.

I tried to leverage existing code (hence the cmove implementation to put the search word to here).

A better design would change match? to refer to a text match buffer instead of here.  Then 'n' and 'p' will be easy to add.


burnsauce:
As I see we're bumping up against the cart size limit shortly, I will rescind this PR because it's a poor implementation and wasteful.

PR #230, opened 2020-02-22
Makefile now launches deploy with nice VICE settings

raydeejay:
Reset to default settings in case drive 8 is missing,
or otherwise misconfigured, and enable warp mode.


jkotlinski:
Thank you! Appreciated!

jkotlinski:
Hi again, FYI I had to remove the -default. My Vice install somehow does not work with default settings.

PR #231, opened 2020-02-24
Filesystem words

raydeejay:
Most of them use DOS to do their work and are safe to use. Some use
the memory past EOF, and therefore both rely on the editor being
loaded and can potentially corrupt the higher parts of memory if used
on big files.

When I figure out how to read and write files block by block I will
improve the implementation.

I accidentally committed a Makefile with this file already added to the list,
and it went over both of us! :-)


jkotlinski:
Hi! Neat code! Seems like something that would fit in nicely.

Some thoughts about module name. Maybe "diskutil" would be more to the point than "filesystem"? And possibly "ls" would fit well in here.

I'm not convinced about the need for namespacing. Instead the word names could be more explicit.
E.g. "fs.copy" is somewhat ambiguous, as it does not tell if the disk or file is to be copied. Same with "fs.rename". If possible, I'd like to have it follow [Forth 2012](https://forth-standard.org/standard/file) standard word naming, where it makes sense.

Also it would be nice to have it documented like other modules.

What do you think?

raydeejay:
All of that sounds reasonable, I'll give it some more thought :+1:

jkotlinski:
Great!
Come to think of it, maybe file.fs is enough. Short, sweet and standard
conforming :-)

raydeejay:
Now I'm confused, would "file.fs" be the name of the forth source file on `forth_src/` ? :thinking:

jkotlinski:
Hi, sorry I was in a rush while writing that.
Yes, I thought the module could be named ”file” rather than ”filesystem” or
”diskutil”.
Just an idea though.

Came to think of it, maybe forget-<module name> would be better name for
the marker?

raydeejay:
I agree that `forget-module` is a very explicit name, but I'm very fond of the `---module---` convention since I picked it up from uhh... somewhere :-D because I like how it makes it stand out when issuing `words`.

I think the only candidate for a standard name in this set of words is `rename-file` :thinking:

raydeejay:
I was just thinking that there's probably not enough space on the cart to fit the whole set of words in, which means `ls` is probably better off being a "builtin", whereas these words can be loaded when needed and then unloaded without prejudice.

jkotlinski:
Hi!

 1) if you like that marker name, I like it too!
 2) it might be that RENAME-FILE (and possibly DELETE-FILE) is only direct
match, but other words could be named in same style, too. I like -FILE and
-DISK suffixes because then it’s 100% clear from word names what is going
on. And e.g. RENAME-DISK and RENAME-FILE can exist at the same time. Of
course, the third option is to go all out Unix and name it MV instead...
but I usually prefer to be Forth-standard-like when possible
 3) about LS being ”in-built”: OK

raydeejay:
Hmmm... what if it was simply called `dos`?

---cut here----------------------------------------
marker ---dos---

`mount  ( -- )` DOS init
`check-disk  ( -- )` DOS validate 

file operations:
`copy-file  ( "src" "dst" -- )` DOS copy, with more natural order of arguments
`rename-file  ( "src" "dst" )` DOS rename
`delete-file  ( "filespec" -- )` DOS scratch
these need to be rewritten soon to copy blocks instead of the whole file at once, so they can deal with files of any size (maybe have the disk drive do the work instead?):
`join-files  ( "src1" "src2" "dst" -- )` join two files in memory and save a single file afterwards
`copy-to-disk  ( "file" dst -- )` copy a file from the current drisk to another

disk operations:
`format-disk <label,id>`
    (to be implemented later):
    `<src> <dst> copy-disk`
    `rename-disk <newlabel>`

---cut here----------------------------------------

A further refinement is to make the words above wrap around these:

marker ---(dos)---

`(copy-file)  ( addr1 n addr2 n -- )`
`(rename-file)  ( addr1 n addr2 n -- )`
`(delete-file)  ( addr n -- )`
`(format-disk)  ( addr n -- )`
`(rename-disk)  ( addr n --)`  (TBI)

I only listed here the words that take string parameters (which justifies having a parsing version for interpreter use and a non-parsing version for compiling) and also use DOS to do the work (not needing a buffer). 

If going this way, two markers could be provided so the interactive-only subset of words can be forgotten after loading. Or split the file in two. Or do nothing, of course xD

jkotlinski:
Sounds great!

raydeejay:
The `(word)` form is rather ugly... I'm thinking of removing the parens on those words, and adding the `fs.` prefix back to the interactive words. Thoughts?

jkotlinski:
The use of `(parentheses)` for internal words is very established! You will find this in FIG-FORTH code from 1983, as well as Gforth source code...

In contrast, namespaces are rarely used in Forth. Word lists are the typical way to avoid naming conflicts - although DurexForth does not have that functionality yet.

raydeejay:
I was playing with ForthOS before durexforth. ForthOS is a not-really standard bare metal x86 forth, with vocabularies, that allows you to refer words from a vocabulary without having it in the search list, using the `voc.word` syntax. That's where I picked it from.

I know that the use of parentheses is very established, but those are the words to be used in word definitions, so I though it would be more standard-like to have the programmatic words resemble the standard, and give a different name to the interactive words.

jkotlinski:
I'm a bit disoriented, since right now I only have the time to look at this project about 15 minutes every third day :-D
But I can elaborate on the word naming strategy for modules.
 * Do not use name spaces
 * Words that are only used internally within a module may optionally be wrapped with parentheses

I.e. if you do not like the parentheses, you do not have to use them. E.g. for your new module, I think `0cmd` and `(0cmd)` are equally fine.

raydeejay:
Life gets in the way, I get it don't worry xD

Given what has been discussed so far, I'm thinking of two alternatives:

- Make the words state-smart, which I really don't like but it would allow to use the same words in interpret and compile mode, but with different syntax (ugh)

or

- Split into two modules: the first defines the words normally for programmatic use. The second redefines the words for interpret/"shell" use, the rationale being that, for most people most of the time, the interactive set of words is what they'll want.

I'm guessing you'd prefer the latter?

jkotlinski:
Sorry, I think I finally understand the problem.
First I thought, why not discard the interpreter versions? But that won't work, since S" is compile-time only.
The standard has e.g. INCLUDE (for interpreter) and INCLUDED (for compiler) but it is hard to come up with a similar naming scheme for these words.
I think I will need to sleep on it ;)

jkotlinski:
Hi again!
I thought about it a little more. Maybe two modules would be preferable
anyway?
Personally I would like to use RENAME-FILE and DELETE-FILE in v, then it’s
good with a minimal module.
The interactive versions I don’t care too much about myself, but I can see
the point of having them available.

raydeejay:
Mmmm I don't know if there will be enough space on the cart to load V on top of the DOS words :thinking:

jkotlinski:
That can probably be fixed. EIther by size optimizations, or by offloading
rarely needed words to compat, or something else :-)

jkotlinski:
For what it's worth, I worked a little on cutting down size, and now there are over 256 bytes free on the cartridge.

raydeejay:
I'll see if things can fit there :-)

jkotlinski:
@raydeejay any updates?

Whammo:
There doesn't seem to be much that couldn't be done with,

: dos parse-name $F openw $F closew ;

 And it's nice to be able to check for success as you go.

: dstat $1710 sys ;

jkotlinski:
Closed due to lack of activity. Please let me know if you want to reopen.

PR #233, opened 2020-02-26
Simple printer driver

raydeejay:
As the label says, a handful of words to get things on paper, virtual or not :-)

I'm pretty sure that something more should be done on that piece of assembly, but I'm not sure what.

### DIFF
--- a/forth_src/printer.fs
+++ b/forth_src/printer.fs
@@ -1,0 +1,59 @@
+\ simple printer driver for a printer
+\ connected to #4
+
+code print  ( addr n -- )
+  \ preserve stack depth
+  w stx,
+
+  \ no filename
+  0 lda,#   0 ldx,#   0 ldy,#
+  $ffbd jsr,
+
+  \ file, device, channel
+  4 lda,#   4 ldx,#   7 ldy,#
+  $ffba jsr,
+
+  \ open
+  $ffc0 jsr,
+
+  \ chkout
+  4 ldx,#
+  $ffc9 jsr,
+
+  \ restore the stack depth and
+  \ simply call 'type'
+  w ldx,
+  ' type jsr,
+  w stx,
+
+  \ clrchn
+  $ffcc jsr,
+
+  \ close file
+  4 lda,#
+  $ffc3 jsr,
+
+  \ TODO: it seems that some more
+  \ cleanup should be done...?
+
+  \ restore stack depth
+  w ldx,
+;code
+
+\ prints the file being edited
+: vprint  ( -- )
+  bufstart eof @ over - print ;
+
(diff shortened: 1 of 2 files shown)


forth_src/printer.fs
```diff
@@ -0,0 +1,59 @@
+\ simple printer driver for a printer
+\ connected to #4
+
+code print  ( addr n -- )
+  \ preserve stack depth
+  w stx,
+
+  \ no filename
+  0 lda,#   0 ldx,#   0 ldy,#
```
polluks: `tax, tay,`
But I wrote `OPEN 4,4`, so it's always ignored.

polluks:
At first you may check if the printer is online, like this:
````
10 REM IEC-GERAET TESTEN
20 POKE 780, GA
30 POKE 144, 0: REM STATUS LOESCHEN
40 SYS 63048: REM SENDE "LISTEN" UND "UNLISTEN"
50 PRINT ST
````

raydeejay:
> At first you may check if the printer is online, like this:
> 20 POKE 780, GA

Uhhh what's `GA`?

polluks:
@raydeejay the device number however kernal calls are better than direct calls.

raydeejay:
Oh, so you can do that... xD I'll give it a test drive

raydeejay:
I see, I'll write some examples instead. Is there a similar mechanism for reading from a file? :thinking:  I'll need it for copying files in blocks.

jkotlinski:
I am not sure how to efficiently read blocks using DOS. Maybe there is a way? I tend to mostly use file operations, as they are sped up nicely by cartridges like Action Replay.

To get the most efficient block operations, maybe you will have to upload custom code to the 1541. I am no expert in this but there is a lot of info to be found: https://codebase64.org/doku.php?id=base:fastloaders
I am not sure if it's more trouble than it's worth.

jkotlinski:
By the way. I think the best way to copy a file is not to write your own code, but rather use drive command.
For 1541, it's like this in BASIC: `OPEN 1,8,15,"C0:NewFile=OldFile": CLOSE 1`

raydeejay:
> By the way. I think the best way to copy a file is not to write your own code, but rather use drive command.

Not if you want to copy across drives... :grin:

jkotlinski:
Oh true! I didn’t think of that... how silly!

jkotlinski:
@raydeejay any updates?

Whammo:
There's copy between drives, then there's copy between devices.

jkotlinski:
I am closing this PR now, since a new printer driver is probably unneeded. I will create a new issue for creating an example on how to do printing with the words already available.

PR #234, opened 2020-02-27
[FR] Show 'position/size' in the editor statusbar

raydeejay:
Submitted for your consideration, 126 additional bytes in the dictionary adding a display bottom right in the editor, showing the current cursor position and the total length of the file (minus the prg header, I guess).

Alternatives:
- `x/100% (size)`
- `line, column` (not that useful without the size)
- `line/lines, column
- also display the number of blocks it takes on disk
- the best of all of the above
- not including it at all

All these will cost more bytes, of course. I measured the size by comparing `here` with and without the addition. How can I tell how many bytes are left free, other than watching the cart build fail? xD

### DIFF
--- a/forth_src/v.fs
+++ b/forth_src/v.fs
@@ -468,6 +468,14 @@ linelen dup if 1- then min
 curx c! then cursor-scr-pos
 dup @ $80 or swap c!
 
+\ show position
+$7e8 >r
+eof @ bufstart - 1- 0 <# #s #>
+r> over - dup >r swap move
+r> 1- '/' over c! >r
+editpos bufstart - 0 <# #s #>
+r> over - swap move
+
 key
 
 \ hide cursor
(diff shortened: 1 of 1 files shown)


burnsauce:
I think the cart build failing is the easiest way to detect a too-large dictionary.

That said, I've mentioned before that the V editor will no longer fit on the cart with more than a few changes.  Maybe the editor should be a stripped down version for the cart with a full-featured editor on the disk?

Kroc:
Is the editor stored as text on the cart? Compression could be used to fit more in

burnsauce:
@Kroc no, the editor is compiled into the dictionary of the cart.

jkotlinski:
I like this, maybe especially the options "line/lines" or "100%".

At this moment though, it seems we are tight on space, so maybe this should wait until more space is freed up?

raydeejay:
Fine with me! :grin:

burnsauce:
More space is freed up :)

raydeejay:
I'm tied up in other things and not really working on durexforth now :-/

jkotlinski:
Soo... finally I got the time and peace of mind to try this out now. I'm happy to see it works as expected, even if some time passed.

I find that byte position + file size is not _that_ helpful. Would it be possible to switch to line and column? Are you interested in working on this further?

Issue #19, opened 2015-03-14
Editor: * find


GoogleCodeExporter:
```
Find word under cursor position
```

Original issue reported on code.google.com by `kotlin...@gmail.com` on 12 Jan 2009 at 4:34

polluks:
Just FYI http://forum.6502.org/viewtopic.php?f=9&t=4364

jkotlinski:
Oh... this ticket is a bit unclear. It is about search in the editor, * command in Vim: http://vim.wikia.com/wiki/Searching

polluks:
I know, I just want to save the interesting link somewhere.
And how about this? http://forum.6502.org/viewtopic.php?f=9&t=4294

PR #235, opened 2020-02-27
Implemented '*' and 'n' with factored search buf

burnsauce:
This branch factors out text searching into a counted string, permitting a repeat of the previous search with the 'n' key.

Additionally, the star '*' key will search for the word under the cursor, fixing issue #19

### DIFF
--- a/forth_src/v.fs
+++ b/forth_src/v.fs
@@ -323,18 +323,37 @@ key case
 'd' of del-line endof
 endcase clear-status ;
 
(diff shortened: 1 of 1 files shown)


burnsauce:
I created a new buffer, fbuf, to replace the use of here pad space for repeated searching and * functionality.

I notice there's a clip buffer, I don't know why it's smaller than the herebuf used was.  I stuck with 38 bytes (39 with the count).  I don't know if this should in fact be smaller.

PR #248, opened 2020-04-13
ls update

Whammo:
It parses the $ instead of keeping count.  You don't have to load it again to have another look. Load a $ anywhere in memory and list. ls is a little bit smaller. (maybe)

### DIFF
--- a/docs/anatomy.tex
+++ b/docs/anatomy.tex
@@ -40,7 +40,7 @@ The name length byte is also used to store special attributes of the word. Bit 7
 The code section contain pure 6502 machine code.
 
 \begin{description}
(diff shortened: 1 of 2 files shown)


Whammo:
Let me sort this out before I submit this. It works well, but my submissions are a mess.

jkotlinski:
As you prefer. By the way, I think it would be nice if it will stay
possible to just type ”ls”, it is very convenient!

Whammo:
Johan,

 

I’m trying to consider different priorities.


Okay, ls will always load a directory here and go out to the screen.




Memory space. Is this an issue?

Separating the low level reading and display routine for other uses. ( possibly dump, certainly inspect )

Re-reading the directory without loading again.

Using wildcards and drive numbers. (  $9: du*=p )
Chaining directories i.e. loading to arbitrary places in memory



Just things I’m finding useful and fun to do.

If you have any thoughts.

 

  Thanks

 

  Kev

PR #254, opened 2020-04-29
Migrate from HERE to PAD

Whammo:
I believe I went through all the .fs files, I was leery of the actual applications including V, so the changes are conservative, and my notes will likely clarify my intentions.
 There are really not that many references to HERE in there. I steered clear of likely compiling words.

### DIFF
--- a/forth_src/base.fs
+++ b/forth_src/base.fs
@@ -85,7 +85,7 @@ msb lda,x lsb sta,x
 dup code lda,# 100/ ldy,#
 ['] pushya jmp, ;
 : constant value ;
-
+$33c constant pad
 : space bl emit ;
 : spaces ( n -- )
 begin ?dup while space 1- repeat ;
--- a/forth_src/float.fs
+++ b/forth_src/float.fs
@@ -15,7 +15,7 @@ s" .8" strf .8
 : bsys \ system call to BASIC ROM
 1 c@ dup 3 or 1 c! swap sys 1 c! ;
 : fac, $bbca bsys
-$57 here 5 move here 5 + to here ;
+$57 pad 5 move pad 5 + to pad ;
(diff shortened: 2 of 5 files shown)


jkotlinski:
The standard says:

"The size of the scratch area whose address is returned by PAD shall be at least 84 characters. The contents of the region addressed by PAD are intended to be under the complete control of the user: no words defined in this standard place anything in the region, although changing data-space allocations as described in 3.3.3.2 Contiguous regions may change the address returned by PAD. Non-standard words provided by an implementation may use PAD, but such use shall be documented.

Rationale:
PAD has been available as scratch storage for strings since the earliest Forth implementations. It was brought to our attention that many programmers are reluctant to use PAD, fearing incompatibilities with system uses. PAD is specifically intended as a programmer convenience, however, which is why we documented the fact that no standard words use it."

So it would be more to the intent of the standard if Durexforth provides PAD, but does not actually use it itself.

Whammo:
I did not (I believe) alter any Forth standard words, and although I might argue for Format.fs, Float.fs, and V.fs implementing non-standard words that use these spaces as a convenience for dealing with strings.
 I also agree upon second thought.  Format.fs and Float.fs, are not applications, but vocabularies. PAD should be not effected by these calls because if the user is in PAD he doesn't want these calls corrupting PAD.
 However, V, being an user space application, ( Although systemic) there is no issue of concurrency in PAD, or HERE. And there is also no reason (other than using and demonstrating the use of PAD) to migrate V from HERE to PAD.
 Agreed that Float.fs and Format.fs shall not use PAD.
I would like to appeal that V as an application should use PAD.
 That said, it's not going to hurt my feelings if denied. 
  It's all fun, thanks guys!

PR #258, opened 2020-05-22
LS responds properly to BLOAD disk errors

Whammo:
LOADB will print an error message and return zero on a failure to load.
LS now checks this, and aborts leaving the error message


jkotlinski:
I wonder if something is wrong with the pull request. It's 20 commits and some seem unrelated. Maybe it should be from some branch instead of Whammo:master ?

Whammo:
Yes, sorry for my ineptitude with Git.

Let me take a look at things and see what I can do.

Whammo:
I will re-submit.

Issue #269, opened 2020-10-12
Shrinking dictionary


polluks:
Something like this https://stackoverflow.com/a/38104617/1430535

burnsauce:
There's a project that does this as a cross-compiler: https://github.com/jkotlinski/acmeforth

Natively, I don't think some of those improvements apply to durexforth core, as it has to be interpreted.

polluks:
I see, thx.

burnsauce:
@polluks looks like this issue is indeed valid!

burnsauce:
I created 2 incompatible pull requests as ideas for moving forward.  The latest is my attempt to remove the dictionary from data space.  This will permit the stripping of the dictionary for turn-key deployment.  It also saves 2 bytes per word ;)

burnsauce:
Now that the dictionary headers have been split from the code, the potential exists to shrink the space used by the headers.  Several options exist now:

- `SAVE-PRG` strips all headers from the saved PRG file, already implemented.
- `HIDE` has been floated to remove individual word headers, easily implemented with `MOVE`.
- `HIDE:`, `;HIDE`, and `SHOW` ( or perhaps `MODULE:`, `;MODULE`, `EXPORT` ) for bulk hiding with selective export.  Example:

```
header foo
hide:
: bar ;
: baz ;
:noname bar baz ; show foo
;hide
```

`bar` and `baz` headers don't appear in the final dictionary, and `foo` is assigned the XT of the `:noname`.

Experiments show that judicious use of these reduction techniques leaves over 500 bytes free on the cart build, currently more than 300 bytes in deficit.

jkotlinski:
How about e.g.

    defer myword
    latest
    : bar ;
    : baz ;
    :noname bar baz ; is myword
    to latest

? then no new words need to be added and I think it's easy to understand.

burnsauce:
That works! Defer isn't in the base distribution, though.

jkotlinski:
I think it is fine, it could be added first thing after `---modules---` master perhaps, then it does not affect minimum strippable size and can be used for the modules.

burnsauce:
Looks like that works out well.  PR submitted.  Now it's just down to stripping the words that aren't required.  Many that I thought would not be are actually useful in `SEE`, e.g.: `(OF)`, etc. so I don't really know what's fair game, like `lstk` etc.

`V` can be stripped of almost everything except `V`, `BUFSTART`, and...?

[...]

PR #271, opened 2020-10-14
Mostly implemented SEARCH-ORDER

burnsauce:
Here is a simple implementation of the SEARCH-ORDER wordset (except for SEARCH-WORDLIST).

As it is an increase in size, the cartridge will not build. (edit: it's only a few bytes off, which can be trimmed by reducing a few internal words to NONAMEs with helper functions in a different wordlist)

I don't expect this to be merged any time soon, if at all.  I just put it out here as an idea, given the discussion about changes to the dictionary.

Enjoy!

edit: WORDS will need SEARCH-WORDLIST to comply with the standard:
"List the definition names in the first word list of the search order."
Currently, it shows the CURRENT wordlist, not the first in the search order.

MARKER will also need an update to backup and restore WIDS, _ORDER, and CONTEXT

### DIFF
--- a/compiler.asm
+++ b/compiler.asm
@@ -121,9 +121,12 @@ SEMICOLON
     +BACKLINK
     !byte	9
     !text	"immediate"
-    lda	_LATEST
+    lda CURRENT
+    asl
+    tay
+    lda	WIDS,y
     sta	W
-    lda	_LATEST + 1
+    lda	WIDS+1,y
     sta	W + 1
     ldy	#2
     lda	(W), y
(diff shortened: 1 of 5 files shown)


burnsauce:
The motivation for this addition is in furtherance of #269, as it appears to be the standard idiom for malleable wordlists, which could be leveraged for code compression and turn-key deployment.

This implementation would not survive the separation of the dictionary from the data space, as proposed by @jkotlinski .

It also includes a kludge to leverage _LATEST instead of rewriting compiler.asm to JSR to an updated LATEST that points to WIDS.

PR #276, opened 2020-10-16
Split Dictionary

burnsauce:
Following up on issue #269 with an architectural change.

This branch strips the dictionary from the data space and creates a new downward-growing dictionary at $6fff, which is constructed so as to be relocatable.

Current status:
* Bug-free?
* Mostly documented.
* API in flux: DEFER vs SHOW?
* Hide SAVE-PACK?
* marker will not function if the dictionary is edited - documented

Old:
* ~~memory map far from ideal.~~ not addressing.
* ~~need new words for navigating the dictionary~~
* ~~see is broken for the last defined word~~
* ~~save-forth will need to be changed to tuck the dictionary in~~
* ~~likewise, start will have to be modified to untuck it~~
* turnkey operation entirely possible by throwing away the dictionary when saving
* ~~there should probably be a word for relocating the dictionary~~ added `TOP!`
* saves 2 bytes per word!

Extra note: There was a comment saying that ACME didn't like strings as macro arguments.  I had no problem compiling this code with ACME 0.97


burnsauce:
`MARKER` is also broken.

Kroc:
ACME 0.97 specifically added strings to macro arguments 👍

burnsauce:
Note:  The dictionary is smaller than default because it's not a linked list, it uses arithmetic of the string length.

It's a little more tedious to navigate the dictionary now, you can't just `LATEST @ @ @ @`.  I notice that `LATEST` is a nonstandard word, so perhaps a new nonstandard word should replace it.  Does `LATEST` only exist now for the purpose of dictionary navigation?  Maybe an indexed idiom would work better here.

Then again, this dictionary structure begs a new set of words to manage it.  It should be trivial to edit words out, rename them, alias them, etc.

burnsauce:
Running `include test` fails with the default `top`, as the dictionary runs out of space.  In 2.0.0, running test stomps the editor buffer, so it's expected.

Using `$cbff top!`, the tests all pass, however.

burnsauce:
This architecture permits zero-overhead `DEFER`.
```
: defer header ;
```
The `IS` implementation simply replaces the XT of the dictionary entry instead of modifying the arguments of a JSR.

jkotlinski:
Wow! This is a really promising change. I will try to follow up with constructive comments.

interpreter.asm
```diff
@@ -114,7 +110,7 @@ interpret_loop
     rts
 
 interpret_tib
-    jsr	INTERPRET
+    jsr INTERPRET
```
jkotlinski, line 113: I think there will need to be a check that detects if code space and header space have grown so big that they are threatening to overwrite each other. It might be that this is the best place to put such a check. I am not sure how big the safety margin should be.

jkotlinski:
For the problem of traversing wordlist, I would suggest implementing [TRAVERSE-WORDLIST](https://forth-standard.org/standard/tools/TRAVERSE-WORDLIST)

interpreter.asm
```diff
@@ -244,98 +232,91 @@ FIND
...
+    ldy #2
+    lda (W), y
+    and #STRLEN_MASK
+    clc
+    adc #3
+    adc W
+    sta W
+    bcc +
+    inc W + 1
++   lda(W), y
```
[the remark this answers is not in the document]
burnsauce: TOP -1 and TOP are set to 0.

durexforth.asm
```diff
@@ -97,13 +97,22 @@ _START = * + 1
...
-    !set	LINK = * - 2
+!set TOP = $6fff
+!set DICTOP = TOP -1
+
+!set BACK = *
+* = DICTOP
+!word 0
+* = BACK
+
+!macro BACKLINK .name , .namesize {
```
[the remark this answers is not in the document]
burnsauce: You can calculate the string length with len() but you still need to include an argument for flags.  I don't think macros can be variadic so I just kept it the same for simplicity.

durexforth.asm
```diff
@@ -163,14 +160,17 @@ ONE
...
 !src "disk.asm"
 
+    +BACKLINK "top", 3
+    +VALUE  TOP
+
 ; LATEST - points to the most recently defined dictionary word.
-    +BACKLINK
-    !byte 6
-    !text	"latest"
+; OBSOLETE!!! No user facing dict lists?
```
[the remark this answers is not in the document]
burnsauce: The wordlist is useful, but `LATEST` is a lot less useful, as it only ever points to the top of the dictionary, and you can't navigate from there easily.  I think there need to be built-in words to manage the dictionary that will give no call for the user to get that address.

durexforth.asm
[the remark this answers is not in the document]
jkotlinski: Ok! I think it might be a nice thing to calculate string length automatically, but I agree it is better to keep such a change outside this pull request.

interpreter.asm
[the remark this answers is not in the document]
jkotlinski: Right - but doesn't this read from TOP -2 ?

interpreter.asm
burnsauce: Also, what code formatting should be in play here?  I thought it was expanded tabs = 4 spaces, but I guess that just happened to align when I retabbed the file.

burnsauce:
> For the problem of traversing wordlist, I would suggest having a look at [TRAVERSE-WORDLIST](https://forth-standard.org/standard/tools/TRAVERSE-WORDLIST). To me it seems like a nice way to solve it.

Cool, that's like a lambda!

If it's going to match the standard, there's either going to be an extraneous argument (`wid`) that gets dropped, or I should look to implement SEARCH-ORDER which seems like a nightmare with this dictionary structure without ALLOC.

Or maybe just nonstandardize it and forget `WID`?

durexforth.asm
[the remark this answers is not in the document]
jkotlinski: I suppose it will remain useful - if nothing else, to figure out start of header space.

interpreter.asm
```diff
@@ -244,98 +232,91 @@ FIND
...
+    cmp #0
+    beq .string_compare
 
 .word_not_equal
-    ; no match, advance the linked list.
-    ldy	#0
-    lax	(W), y
-    iny
-    lda	(W), y
+    ; no match, advance the dp
```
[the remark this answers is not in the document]
burnsauce: The other advantage is that removing the linked list allows the dictionary to be portable.  Is compile / interpret time a critical goal that shouldn't be overlooked?

interpreter.asm
jkotlinski: It is a good question. I also think it should be expanded tabs=4 spaces. I have not followed that consistently in the past. I could make an issue to add an [.editorconfig](https://editorconfig.org/) file for clarity.

interpreter.asm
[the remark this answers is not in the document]
burnsauce: Oh you're right!  I should be looking at the XT field there, or extend TOP another byte?

interpreter.asm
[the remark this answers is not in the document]
jkotlinski: It is something to keep in mind. In the past, I have timed how long it takes to make a full build, trying to keep it not more painfully slow than needed :)
For now, I suggest no changes to this code for the sake of speed. Once everything is in place and the wordlist is stripped from useless entries, it will be very interesting to again measure the full build time.

jkotlinski:
Since wordlists are not implemented, it seems like the right thing to implement it without `wid`.

interpreter.asm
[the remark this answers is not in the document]
burnsauce: Fixed.

durexforth.asm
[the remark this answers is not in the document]
burnsauce: We could save 2 bytes and a bunch of JSRs if LATEST were to be a value instead of a variable.  Given that the dictionary access idiom is changing, can LATEST change?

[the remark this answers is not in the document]
jkotlinski: I am not sure. It would be an API change, meaning major version bump. In practice though, a change like this will likely break some code due to the change in word structure, so a major version bump could be warranted anyway. What do you think?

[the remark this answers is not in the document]
burnsauce: My reading of semantic versioning says yes, my gut says it's funny to release another major version so soon.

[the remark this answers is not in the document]
jkotlinski: I think it has to be yes :-)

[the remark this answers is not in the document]
burnsauce: In that case, LATEST is now a value.

interpreter.asm
burnsauce: So, if I do a check for unused space right here, won't a full dictionary result in a deadlocked interpreter?

forth_src/v.fs
```diff
@@ -568,3 +570,8 @@ rom-kernal bufstart loadb
 ?dup 0= if reset-buffer else
 eof ! 0 eof @ c! then
 else drop then main-loop ;
+
+swap 
+to latest
+: v ;
+latest !
```
jkotlinski, line 577: I think it would be nice to have something like 

    MODULE
    ...
    : v ... ; PUBLIC
    ....
    ;MODULE

where `;MODULE` strips all words within the module that are not tagged as `PUBLIC`.

Also it would make sense to bring back good old `HIDE` , to remove words from the wordlist.

burnsauce: Would `HIDE` set the hidden flag or edit the dictionary exactly like `DELETE`?  If the latter, MARKER is going to need a rework.

jkotlinski: There are other ways to do it, too...

    MARKER ---module---
    ...
    : V ... ; SHOW
    ...
    HIDE-TO ---marker---

interpreter.asm
jkotlinski: I think it would make sense to print warnings when there is, for example, less than 256 bytes left.

forth_src/v.fs
jkotlinski: Humm... I didn't see `DELETE` until now, to be honest. I think actually what I mean is to rename `DELETE` to `HIDE`.

forth_src/dict.fs
```diff
@@ -0,0 +1,11 @@
+: delete
+parse-name find-name 
+if
+xt> ( dp )
+dup 2+ c@ 3 + ( offset )
+swap latest - >r ( offset ) 
+latest + ( dst )
+latest swap dup to latest ( src dst ) r>
+move exit then 
+type ."  not found"
```
jkotlinski, line 10: `ABORT"` is pretty nice for things like this.

durexforth.asm
```diff
@@ -97,13 +97,22 @@ _START = * + 1
...
+!byte 0 ; we need this byte here to mark TOP without slowing FIND-NAME
+* = BACK
+
+!macro BACKLINK .name , .namesize {
+    !set DICTOP = DICTOP - 3 - len(.name)
+    !set .xt = *
+    * = DICTOP
+    !word .xt
+    !byte .namesize
+    !text .name
```
jkotlinski, line 114: I think it would make sense to change to

    .flags/namesize
    .name
    .xt

The most important operation on the wordlist is word finding. Then it becomes easier to understand if data is accessed by the order it is accessed. First flag byte, then name, then finally xt. This will also allow for having a single sentinel 0 byte at dictionary top.

burnsauce: Hilarious.  That was my first design, but I switched it around for ease of finding XTs.

I'll find the commit and switch it back around.

jkotlinski:
Haha! You shouldnt take my comments as the law :-) maybe there was a good
reason to have it in this order that I’m not seeing...

durexforth.asm
jkotlinski: I'm not sure what you mean with ease of finding xt's..?
"FIND wordname" should return the xt ... but I'm probably missing something :-)

jkotlinski:
About your question "Does LATEST only exist now for the purpose of dictionary navigation?"

I think `LATEST` can be used for wordlist traversal but another one is like `IMMEDIATE` uses it.

forth_src/v.fs
jkotlinski: I think it's not worth having `DELETE`/`HIDE` update markers. It is fine leaving it unsupported as long as it is mentioned in the manual.

burnsauce:
> About your question "Does LATEST only exist now for the purpose of dictionary navigation?"
> 
> I think `LATEST` can be used for wordlist traversal but another one is like `IMMEDIATE` uses it.

Immediate is in assembly though, so it doesn't need `LATEST` in the dictionary.

interpreter.asm
burnsauce: Doing it right there will be either a lot of code or will use some stack space, requiring the stack check to be shallower, I think.

interpreter.asm
```diff
@@ -244,62 +232,50 @@ FIND
     jsr R_TO
     jmp ZERO
 
-    +BACKLINK
-    !byte	9
-    !text	"find-name"
+    +BACKLINK "find-name", 9
 FIND_NAME ; ( caddr u -- caddr u 0 | xt 1 | xt -1 )
```
burnsauce, line 236: I think maybe this should be refactored to use `TRAVERSE-WORDLIST`.

burnsauce: Actually that would slow it down, nevermind.

interpreter.asm
```diff
@@ -110,9 +110,13 @@ interpret_loop
     rts
 
 interpret_tib
-    jsr	INTERPRET
+    jsr INTERPRET
     cpx #X_INIT+1
     bpl .on_stack_underflow
+    lda LATEST_MSB
+    sec
```
burnsauce, line 117: Should I ignore the `sec` here?  Shouldn't make a difference.

jkotlinski: can't it be replaced by cmp?

burnsauce: Oh yeah!

burnsauce: This check isn't good enough, as HERE and LATEST can be on different pages with 1 byte unused.  I'll have to re-do it. (Also it's in the wrong place!)

forth_src/v.fs
burnsauce: Working on this and trying to implement the `SHOW` element, and I'm finding myself in need of scratch space.  By definition, I shouldn't be using `PAD`.  Would it be safe to use $a000 for the entire span of `MODULE` to `;MODULE`?  I.e.: is it okay to add the restriction "the RAM block at $a000 is unavailable for use when compiling modules." ?

burnsauce:
I'm finding a second need for the bit of code in `SEE` that finds the previous word name token, and as you say, naming is hard.  I'm thinking `PREV` or something.  For that matter, the bit of code that advances a name token could be `NEXT` or something.  Thoughts?

burnsauce:
Since we're shooting 3.0.0 here, I was thinking of a revamp of the memory map.

If we say that `TOP` - $cbff is a downward-growing heap, we can `ALLOCATE` a buffer for the editor.  Also then we'd have `ALLOCATE`, `FREE`, and `RESIZE`.

burnsauce:
ALLOCATE (219 bytes) and FREE (131) might be a lot smaller if written in forth instead of assembly.  Performance or code space for these words?

jkotlinski:
IMO, the optimal size for ALLOCATE, FREE and RESIZE is 0 bytes ;-)

Is there a particular problem that calls for the addition of these words?

burnsauce:
MODULE need an editing buffer to enable SHOW.  Either ALLOCATE or stomp $a000, which is fine as long as it remains untouched from MODULE to ;MODULE

burnsauce:
Clearly defined:

For an indeterminate number of words, temporarily store information about either hidden or shown words so that the dictionary can be modified when the module definition concludes.

docs/words.tex
```diff
@@ -367,6 +382,19 @@ \section{System State}
 : x ;
 forget
 \end{verbatim}
+
+\item[traverse-wordlist ( xt -- )] Executes \texttt{xt} for successive headers starting from \texttt{latest}. Each time \texttt{xt} is called, it must consume an argument from the stack and leave behind a truth value. If true (not 0), it will continue to the next header address and call \texttt{xt} again. If false, or if it reaches the end of the dictionary, traverse-wordlist will return. Example:
```
jkotlinski, line 386: Idea!
Since `TRAVERSE-WORDLIST` is anyway not standard, how about calling it something shorter instead, that fits better on screen?
How do you like `DOWORDS`?

forth_src/v.fs
jkotlinski: You know, if my idea leads to such complications, I really think it's a terrible one!

Maybe just doing something real simple instead? E.g.

    MARKER ---EDITOR---

    DEFER V
 
    LATEST @
    ...
    :NONAME ... ; IS V
    ...
    LATEST !

Then no new implementation is needed, besides moving some words out of compat.fs.
In practice, I think this is really a practical and user-friendly solution, because the "public" words are placed in top of the file.

jkotlinski: It has to be acknowledged, this solution is not optimal size-wise because defer will waste some byte or two, but in this case I think the user-friendliness outweights that drawback.

forth_src/test.fs
```diff
@@ -18,7 +18,7 @@ parse-name mmldemo included
 1 fakekeys \ exits demo
 parse-name spritedemo included
 
-latest ! to here
+to latest to here
```
jkotlinski, line 21: I am not sure why this is not using `MARKER`

forth_src/v.fs
burnsauce: Instead of DEFER v, how about HEADER v?  Then no overhead.  Need a word to get the XT address of a word, but that's it.

jkotlinski: Actually, wasn't `DEFER` zero overhead anyway?

docs/tutorial.tex
```diff
@@ -140,12 +140,27 @@ \subsection{Custom Start-Up}
...
+
+If you have developed a program that has no further need of the interpreter, you can eliminate the dictionary entirely before calling \texttt{save-forth} with \texttt{ top to latest }. Now, the header space is 1 byte at \texttt{top}.
+
+As noted above, without calling \texttt{restore-forth}, \texttt{top} will remain 32 bytes above \texttt{here}, or else it will move to its original \texttt{top}. If you wish to keep the header space safe, reserve a byte that remains at value 0. For example, after you have assigned you \texttt{start} word:
+
+\begin{verbatim}
+here 0 , top to latest top! save-forth @0:turnkey
+\end{verbatim}
+
+This allows your program to use memory down to \texttt{here}.
```
jkotlinski, line 162: You know, all of this stuff seems maybe just too complicated.

It should really be enough with two words, `SAVE-FORTH FILE` versus `SAVE-PRG FILE STARTWORD` (or maybe `START-EXE` or whatever?). The system should be smart enough to take care of all the rest.

I understand the idea of compacting wordlist and word definitions for a smaller file. In practice though, is it really worth the extra complication? These days, most people anyway use fastloaders.

burnsauce: The dictionary shrink came as a necessary feature for building the cart.  I like the idea of separating it into distinct words, the cart (and, by proxy, the release disk) created with `SAVE-PRG`.  I'll put this together.

jkotlinski: Oh - what I meant with `SAVE-PRG` is that it would be the strip-wordlist "turnkey" version.
For the cartridge, and the release disk, we would still need the wordlist.

interpreter.asm
```diff
@@ -741,3 +723,55 @@ OLD_BASE = * + 1
...
     !byte 0
+
++BACKLINK "unused", 6
+UNUSED
+    jsr LATEST
+    jsr HERE
+    jmp MINUS
+
++BACKLINK "dowords", 7 ; ( xt -- )
+    ; per the standard, nothing must be left on stack before execute
```
jkotlinski, line 734: the comment is outdated - because it's not standard - but i think the behavior is valuable and it would be good to mention in docs, if not already

jkotlinski: I just happened to remember, I already implemented `TRAVERSE-WORDLIST` a while ago, but reverted it since I didn't have any real use for it:

```
    +BACKLINK
    !byte 17
    !text "traverse-wordlist"
TRAVERSE_WORDLIST ; ( xt -- )
    lda LSB,x
    sta .xt
    lda MSB,x
    sta .xt + 1
    inx
    jsr LATEST
-   jsr FETCH
    lda MSB,x
    beq .ret
    pha
    lda LSB,x
    pha
.xt = * + 1
    jsr PLACEHOLDER_ADDRESS
    lda MSB,x
    ora LSB,x
    bne +
    pla
    pla
    jmp .ret
+   pla
    sta LSB,x
    pla
    sta MSB,x
    jmp -
.ret
    inx
    rts
```

burnsauce: It's not only required, it's necessary.  XT must be able to rely on no distance between NT and the rest of the stack of it's going to use it for reference or return values.

As to that code, I'll adapt the self modifying bit in but as headers are no longer in a linked list, advancing the dictionary will be the bulk of the code.

docs/tutorial.tex
burnsauce: I have implemented `SAVE-PRG` without any wordlist stripping and left instructions about it in the manual.  I can factor out the part that's relevant for the initial build and automate wordlist stripping into `SAVE-PRG`, it's probably a good idea.

forth_src/v.fs
burnsauce: I realize that `DEFER` modifying a `JSR` target has different behaviour than (what I'm calling) `DEFINES`.

If you get the execution token of a `DEFER`red word, you can call it multiple times and get different results if the deferred word changes.  If you get a `DEFINES`-ed execution token, it points to whatever code was assigned to the word at the time you got the XT, it will never change.

At least, I think that's the difference.

forth_src/base.fs
```diff
@@ -106,8 +106,37 @@ then (to) ; immediate
...
+drop swap
+begin ( caddr caddr u )
+1- 2dup + c@ ( caddr caddr u c )
+over 4 pick + c@ <> if ( caddr caddr u )
+2drop drop 1 exit then
+dup 0= if 2drop drop 0 exit then again ;
+
+: name>xta count $1f and + ;
+
+: (string>name) ( caddr u nt ) ( nt 0 | caddr u )
```
burnsauce, line 131: These little lambda words used with `dowords` are probably reusable enough to be documented.

docs/tutorial.tex
burnsauce: I added SAVE-PACK as the third option, in case anyone doesn't have a fastloader ;)

I can hide it if you don't think it's useful in the dictionary.  Speaking of, should I include `HIDE` in base?

forth_src/test.fs
burnsauce: Fixing this uncovered a bug in `MARKER`.  Good test harness :)

forth_src/v.fs
burnsauce: `DEFINES` could be called `SHOW`, then,

```
: module: latest ;
: module; to latest ;
```

Which is dictionary bloat but syntactic sugar.

interpreter.asm
```diff
@@ -244,103 +250,90 @@ FIND
     jsr R_TO
     jmp ZERO
 
-    +BACKLINK
-    !byte	9
-    !text	"find-name"
+    +BACKLINK "find-name", 9
```
burnsauce, line 253: The proposal for `FIND-NAME` returns a name token, not an execution token.  With the new dictionary structure, it's useful to get the name token for a bunch of reasons.  Right now, those needs are supplied by `DOWORDS`, but it's a roundabout, unclear, and code and cycle heavy way of doing it when `FIND-NAME` skips this token on the way to `>XT`.

I think it would make sense to change the signature to return a name token per the proposal, then refactor `FIND` to make the `>XT` call.

burnsauce: After experimentation, I don't think this refactor is worth it.  `HIDE` and `SHOW` will just be bigger words than needed.  If we need code space this could be revisited.

forth_src/v.fs
jkotlinski: Wouldn't this `DEFER` implementation work just fine?

    : DEFER HEADER IS ABORT ;

jkotlinski: I'm tired and I can't code or think much :)  But wouldn't it be enough that `DEFER` creates a wordlist entry using `HEADER` and then sets the xt to `ABORT`?

jkotlinski: ...conversedly, all that `IS` and `DEFER!` need to do is to update xt of a wordlist entry?

burnsauce: > If you get the execution token of a DEFERred word, you can call it multiple times and get different results if the deferred word changes. If you get a HIDE-n execution token, it points to whatever code was assigned to the word at the time you got the XT, it will never change.

I guess the question is: do you ever expect that a user may want to `'` a deferred word in expectation that they can keep that token, `IS` the word, then call that token and get the new value?

```
defer x
: foo [`] x ;
x is bar \ foo execute will run bar
header y
: baz [`] y ;
y show bar \ baz execute will run abort"
```

forth_src/base.fs
```diff
@@ -106,8 +106,55 @@ then (to) ; immediate
...
+latest swap dsize 2dup - to latest 
+over to top swap over - swap 1+ move ;
+
+\ first pass unfactored code ahead
+\ basic dictionary stuff needs to be in base
+
+: name>string ( nametoken -- caddr u )
+count $1f and ;
+
+: (xt>) ( xt1 nt -- nt xt 0 | xt1 1 )
```
burnsauce, line 124: This should not be variadic, as the search can fail.

forth_src/v.fs
burnsauce: The more I think about it, the more SHOW behaviour (modifying the XT of a dictionary entry) is distinct from DEFER/IS.  Consider:
```
defer foo
: bar foo ;
... some code later ...
baz is foo
```
This pattern is typical of `DEFER`, you declare a word so that you can reference it before the behaviour is defined.  If `foo`'s XT was `abort`, baz would compile to a call to abort.  The indirection is essential.

When exporting from a hidden space, the goal is not to reference the word before its behavioural definition, it is to alias some hidden code to a public word for later reference.  While the below code could use `DEFER` and only incur overhead, the above code could not use `SHOW`.
```
header foo
hide:
: bar ;
: baz bar ;
` baz show foo
;hide
```

jkotlinski:
Hi!
I was thinking a bit on how to conclude this PR.
It has been going on for a little while, and we had many ideas good and bad :)  I imagine we could keep coming up with ideas for a good while further.
I think to wrap up this PR effectively, maybe we could backtrack a bit and make only the core of the change? That is, the splitting of dictionary, with the minimum of changes and code to make it work. Then we could take it from there and open a new PR for each further improvement. I was thinking the process might be more manageable that way, and code will reach master sooner. How does that sound to you?

burnsauce:
Sure. I'll put together the minimum.

PR #281, opened 2020-10-27
Minimal Split Dictionary, redux

burnsauce:
Here is an even smaller version of the split dictionary changes.  It is functional except for the cartridge, which requires 212 bytes to be removed from the dictionary.

I have taken care to try to create the smallest possible set of diffs, hence the least number of changes.  There are also changelog and documentation fixes.

I'm not going to close the other one in case you prefer.  Again, I am not in a hurry and do not this change in master any time soon.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,16 +9,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
  - LS now accepts wildcards, drive #'s or not.
  - Dictionary restructured, now header and code data are split.
- - Header data is not a linked list, and grows downward from a default TOP of $6fff. Record structure: `len_flags | str | xt`
+ - Header data is not a linked list, and grows downward from $9fff. Record structure: `len_flags | str | xt`
  - Prompt displays `ful` when there is less than 256 bytes of dictionary space left.
 ### Added
  - RDIR will display directory formatted data anywhere in memory.
  - PAD Scratch pad memory set to cassette buffer. Untouched by DurexForth
+ - DOWORDS, which allows executing an xt for every word in the wordlist 
+ - Turn-key operation utilities in TURNKEY:
  - SAVE-PACK packs the dictionary together before saving, which unpacks at runtime.
  - SAVE-PRG removes the dictionary and saves the program
  - TOP returns the address of the last byte of the header structure. The value at this address will always be 0.
  - TOP! can be used to specify the position of header data.
- - DOWORDS can be used to scan the header records.
 ### Fixed
  - V did not compile in DECIMAL mode.
  - V long line
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,9 +8,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
  - LS now accepts wildcards, drive #'s or not.
+ - Dictionary restructured, now header and code data are split.
+ - Header data is not a linked list, and grows downward from a default TOP of $6fff. Record structure: `len_flags | str | xt`
+ - Prompt displays `ful` when there is less than 256 bytes of dictionary space left.
 ### Added
  - RDIR will display directory formatted data anywhere in memory.
  - PAD Scratch pad memory set to cassette buffer. Untouched by DurexForth
+ - SAVE-PACK packs the dictionary together before saving, which unpacks at runtime.
+ - SAVE-PRG removes the dictionary and saves the program
+ - TOP returns the address of the last byte of the header structure. The value at this address will always be 0.
+ - TOP! can be used to specify the position of header data.
+ - DOWORDS can be used to scan the header records.
 ### Fixed
  - V did not compile in DECIMAL mode.
  - V long line
--- a/docs/memmap.tex
+++ b/docs/memmap.tex
@@ -13,9 +13,9 @@
 \item[\ldots]
 \item[\$801 - \texttt{here}] Forth Kernel followed by code and data space.
 \item[\ldots]
-\item[\texttt{latest} - \texttt{top}] Dictionary. Grows downwards as needed. Default \texttt{top} is \$6fff. 
 \item[\texttt{bufstart} - \texttt{eof}] Editor text buffer. Grows upwards as needed. Default \texttt{bufstart} is \$7000.
 \item[\ldots]
+\item[\texttt{latest @} - \$9fff] Dictionary. Grows downwards as needed.
 \item[\$a000 - \$cbff] Free RAM.
 \item[\$cc00 - \$cfff] Hi-res graphics colors.
 \item[\$d000 - \$dfff] I/O area. 
--- a/docs/memmap.tex
+++ b/docs/memmap.tex
@@ -11,8 +11,9 @@
 \item[\$200 - \$258] Text input buffer.
 \item[\$33C - \$3FB] PAD Scratch pad memory, Cassette Buffer, untouched by DurexForth. 
 \item[\ldots]
-\item[\$801 - \texttt{here}] Forth Kernel followed by dictionary.
+\item[\$801 - \texttt{here}] Forth Kernel followed by code and data space.
 \item[\ldots]
+\item[\texttt{latest} - \texttt{top}] Dictionary. Grows downwards as needed. Default \texttt{top} is \$6fff. 
 \item[\texttt{bufstart} - \texttt{eof}] Editor text buffer. Grows upwards as needed. Default \texttt{bufstart} is \$7000.
 \item[\ldots]
 \item[\$a000 - \$cbff] Free RAM.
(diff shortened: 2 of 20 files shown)


CHANGELOG.md
```diff
@@ -8,9 +8,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+ - Header data is not a linked list, and grows downward from a default TOP of $6fff. Record structure: `len_flags | str | xt`
+ - Prompt displays `ful` when there is less than 256 bytes of dictionary space left.
 ### Added
  - RDIR will display directory formatted data anywhere in memory.
  - PAD Scratch pad memory set to cassette buffer. Untouched by DurexForth
+ - SAVE-PACK packs the dictionary together before saving, which unpacks at runtime.
+ - SAVE-PRG removes the dictionary and saves the program
+ - TOP returns the address of the last byte of the header structure. The value at this address will always be 0.
+ - TOP! can be used to specify the position of header data.
+ - DOWORDS can be used to scan the header records.
```
jkotlinski, line 21: maybe - "DOWORDS, which allows executing an xt for every word in the wordlist"?

docs/words.tex
```diff
@@ -367,6 +378,14 @@ \section{System State}
 : x ;
 forget
 \end{verbatim}
+
+\item[dowords ( xt -- )] Executes \texttt{xt} for successive headers starting from \texttt{latest}. Each time \texttt{xt} is called, it must consume an argument from the stack and leave behind a truth value. If true (not 0), it will continue to the next header address and call \texttt{xt} again. If false, or if it reaches the end of the dictionary, dowords will return. Example:
```
jkotlinski, line 382: I think this text could be clearer. As an example, from this text I am missing information how to implement `xt`, like what arguments it will get passed on the stack. Another example, "return a flag" would be more to the point than "leave behind a truth value".

Maybe [TRAVERSE-WORDLIST](https://forth-standard.org/standard/tools/TRAVERSE-WORDLIST) could be used as a starting point? Just remove the  the parts that are not relevant or too much information.

durexforth.asm
```diff
@@ -97,13 +97,22 @@ _START = * + 1
...
 ; ----------- macros
 
-!set LINK = 0
-
-!macro BACKLINK {
-    ; it's tempting to add the string as a macro parameter,
-    ; but this does not seem to be supported by ACME.
-    !word	LINK
-    !set	LINK = * - 2
+!set TOP = $6fff
```
jkotlinski, line 100: I think it would be ideal to use the area below BASIC ($9fff) as dictionary top. It is unfortunate that the text editor buffer is in the way.

forth_src/base.fs
```diff
@@ -106,8 +106,31 @@ then (to) ; immediate
 : 2drop ( a b -- )
 postpone drop postpone drop ; immediate
 
+top value oldtop
+start @ value oldstart
+
+: top! ( addr -- )
+latest @ swap top latest @ - 
+2dup - latest ! 
+over to top swap over - swap 1+ move ;
```
jkotlinski, line 115: Having new words added to base.fs looks like somewhat of a drawback to me. I would like to see the minimum RAM footprint of Durexforth smaller than it is today, so anything added (while the impact may not be great) is a tiny step in the wrong direction.

For the new words: SAVE-PRG, SAVE-PACK, RESTORE-FORTH, DSIZE, TOP and TOP! I think the goal should be to remove them, or move them to optional modules. If it's not possible or easily done, then I can live with that :-)

forth_src/test.fs
```diff
@@ -1,6 +1,7 @@
 : fakekeys ( n -- ) $c6 c! ;
 
-here latest @
+$cbff top! \ this section compiles past $6fff.
```
jkotlinski, line 3: I think it's an ugly wart here, that the v edit buffer sits as $7000 and grows unbounded.

Long term, I would like to see v replaced with a much simpler text editor that uses a fixed amount of RAM for the edit buffer, I think ideally no more than $1000 bytes, maybe even $400 would be fine (the number of characters that will fit on a screen). Personally I do not think it makes sense to edit huge text files on C64 anyway.

There are many interesting ideas for interactive Forth text editing historically, so it's pick and choose.

interpreter.asm
```diff
@@ -741,3 +723,48 @@ OLD_BASE = * + 1
...
+    clc
+    adc #3 ; guaranteed carry clear
+    adc .dowords_nt
+    sta .dowords_nt
+    lda .dowords_nt + 1
+    adc #0
+    sta .dowords_nt + 1
+    jmp .dowords_lambda
+; using a word here in case the lambda trashes Ws
+.dowords_nt
```
jkotlinski, line 769: what does dowords_nt mean? i suppose nt is an acronym? could just as well spell it out, we can afford the extra characters when cross-compiling ;-)

durexforth.asm
```diff
@@ -97,13 +97,22 @@ _START = * + 1
...
+!set DICTOP = TOP
+
+!set BACK = *
+* = DICTOP
```
jkotlinski, line 104: There was a comment earlier about naming. Since the dictionary stack grows downwards in memory, isn't it so that what is currently named TOP is actually the bottom? But I guess it depends if one view it as a stack or not. Naming is the hardest part!

docs/words.tex
```diff
@@ -380,6 +399,8 @@ \section{Disk I/O}
 \item[saveb ( start end filenameptr filenamelength -- )] Save binary block.
 \item[device ( device\# -- )] Switches the active device.
 \item[save-forth filename ( -- )] Saves the forth to the given filename.
+\item[save-pack filename ( -- )] Saves a compact version of forth to the given filename.
```
jkotlinski, line 402: Could add a reference to "turn-key operation" section which explains in further detail what actually happens. Same for save-prg.

docs/words.tex
```diff
@@ -356,9 +356,20 @@ \section{System State}
 
 \begin{description}
 
-\item[latest (variable)] Position of latest defined word.
 
-\item[here (variable)] Write position of the Forth compiler (usually first unused byte of memory). Many C64 assemblers refer to this as program counter or \texttt{*}.
+\item[latest (variable)] Address of latest defined header.
```
jkotlinski, line 360: I suppose "header" is now actually a little misleading, since the word is split there is no longer a head or tail. "Wordlist entry"? But it is splitting hairs, we still have the word HEADER , so it keeping it is OK I guess.

jkotlinski (approved):

jkotlinski:
Great job - approved - added some comments that you can deal with as you feel like :-)

forth_src/base.fs
burnsauce: SAVE-PRG can be moved, but SAVE-PACK is essential for both it and the build process.  It could be converted to a build-time-only module at a complexity cost.  As the deploy binary is packed, RESTORE-FORTH needs to remain in functionality if not name.  RESTORE-FORTH requires TOP and TOP!

I could see a RESTORE-FORTH that restores a marker before all of these words to purge them out at first boot.  I think I could make these (eventually) disappear, yeah.  I'll try that unless you prefer otherwise.

DSIZE is already gone, agreed that it was a waste.

burnsauce (commented):
I checked this over, and all is good, as far as I can tell.

docs/memmap.tex
```diff
@@ -11,10 +11,11 @@ \chapter{Memory Map}
 \item[\$200 - \$258] Text input buffer.
 \item[\$33C - \$3FB] PAD Scratch pad memory, Cassette Buffer, untouched by DurexForth. 
 \item[\ldots]
-\item[\$801 - \texttt{here}] Forth Kernel followed by dictionary.
+\item[\$801 - \texttt{here}] Forth Kernel followed by code and data space.
 \item[\ldots]
 \item[\texttt{bufstart} - \texttt{eof}] Editor text buffer. Grows upwards as needed. Default \texttt{bufstart} is \$7000.
 \item[\ldots]
+\item[\texttt{latest @} - \$9fff] Dictionary. Grows downwards as needed.
```
jkotlinski, line 18: Looks fine to me.
I wonder where editor buffer should go in the long term. I was wondering if it's possible to move hires graphics so that it sits beneath Basic ROM, then $e000-$ffff could be used for editor buffer. But it's a tricky question and maybe no great answers.

jkotlinski (approved):
Approved once again :-)

docs/memmap.tex
burnsauce: `HERE $2000 ALLOC 1+ bufstart !`

I mean, does it have to be a fixed address?

jkotlinski: :)  That's a good point. I guess the problem with putting it inside the word space has been that once you get a buffer overrun, it would crash pretty fast.

burnsauce: One more sentinel won't break the bank. ;)

jkotlinski:
Its not so easy because you want the entire file loaded in one go, for the
fastloaders to do its job, and its a bit of work i guess to check in
advance if the file will fit in a buffer.

burnsauce:
> Its not so easy because you want the entire file loaded in one go, for the fastloaders to do its job, and its a bit of work i guess to check in advance if the file will fit in a buffer.

Well, it's a problem for another day then :)

PR #282, opened 2020-10-29
FIND-NAME now return a name token per proposal

burnsauce:
The rationale for this change is this:

The new dictionary structure eliminates the relationship between an XT and its header.  It's no longer a question of simple arithmetic to get one from the other.  DOWORDS exists for scanning the headers, yes, but it's essentially replicating the work FIND-NAME did in the first place, and it's complex and slow.

This change increases the binary size slightly, but reduces the code size and increases the speed of any word that seeks to edit the dictionary. NAME>STRING and >XT would be meaningful outside of a DOWORDS lambda.


jkotlinski:
Makes sense! Thank you :-)

PR #283, opened 2020-10-29
Moved DEFER, etc. to base, after modules marker

burnsauce:


burnsauce:
I realize that this solution is no good.  Users can never get this word back after stripping.  Maybe `include defer` and tack it in its own file?

jkotlinski:
Yes, it makes sense to have `defer` et al in a separate file.

burnsauce (commented):
Forgot to update my master, whoops!

jkotlinski:
An idea about implementation. How about changing `defer` so that xt of the newly created word will be identical to `abort`?
`is` could then simply modify the xt in the word header.

It will be difficult to implement `defer!` this way, since it is supposed to have the signature `( xt1 xt2 -- )`, but for me it is quite fine to remove `defer!`.

jkotlinski:
...maybe my idea above is stupid. But what is anyway the point of `defer!` and `defer@`? Honestly I never used these words so I'll be the first to admit maybe I don't know what I'm talking about :-)

burnsauce:
I don't get the point of `defer!` and `defer@` either.

As to the functionality difference of what's implemented and a dictionary routine, consider:
```
defer foo
: bar foo ;
... some code later ...
baz is foo
```
This pattern is typical of `DEFER`, you declare a word so that you can reference it before the behaviour is defined.  If `foo`'s XT was `abort`, bar would compile to a call to abort, not foo.  The indirection is essential.

jkotlinski:
Oh! You're right, I'm stupid... then `DEFER` is a suboptimal solution after all. It's not very bad, but still. It's a waste of two bytes per word? So one would not want to use it too much.

Another thing that could be added instead of `DEFER` could be `REDEFINE` from Jupiter Ace. Then the same technique would look like:

    marker ---v---
    header v

    latest \ internals
    [ ... all the code ... ]
    : v
    [ ... actual v implementation ... ]
    ; redefine v
    to latest

What do you think? To me it looks kind of neat.

jkotlinski:
Also: Imagine how nice the first page of gfx module would look?

    marker ---gfx---
    header hires \ ( -- ) Enter hi-res graphics mode.
    header lores \ ( -- ) Switch back to text mode.
    [ ... ]

jkotlinski:
Hi again :-)  I read up a little on Jupiter Ace `REDEFINE` and found maybe it's more capable than needed for this, for example, it's able to rename words. Another option might be to add `xt!`, like

    :noname
    \ actual v implementation goes here
    ; defines v

jkotlinski (approved):
Hi again!
Weighing all options, I think it is fine still to go with `defer` for now.
One undeniable benefit is that `defer` and `is` are standard words, and easy to understand for anyone who looks at the code. So it is certainly not all bad.
It would also not prevent a more optimized solution in the future, if one is to change minds.

PR #284, opened 2020-10-30
Shrunk V with DEFER. Fixed bufstart in memory.

burnsauce:
There were no instructions as to how to change `bufstart` properly, so I
didn't bother exporting it.


burnsauce:
Oh yeah also the cart builds now :)

jkotlinski (approved):
Great!
Long-term, it would be great to document how to change the memory layout. Including things like wordlist start, editor buffer, hi-res graphics buffer location.

PR #286, opened 2020-10-30
Added HIDE to defer.fs

burnsauce:
There's probably a better place to put it, but `HIDE` bridges an important
gap between using deferred words to hide words en masse and just needing
to delete a single private word.

It is an uncomplicated way to remove any kind of word (deferred variables / values are a bit messy?) from the dictionary with no overhead penalty.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  - SAVE-PRG removes the dictionary and saves the program
  - TOP returns the address of the last byte of the header structure. The value at this address will always be 0.
  - TOP! can be used to specify the position of header data.
+ - HIDE removes words from the dictionary.
 ### Fixed
  - V did not compile in DECIMAL mode.
  - V long line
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  - SAVE-PRG removes the dictionary and saves the program
  - TOP returns the address of the last byte of the header structure. The value at this address will always be 0.
  - TOP! can be used to specify the position of header data.
+ - HIDE removes words from the dictionary.
```
jkotlinski, line 27: In standard terminology, dictionary is both code+data+wordlist.

maybe "HIDE removes a word from the word list, while leaving its definition in place."

forth_src/defer.fs
```diff
@@ -5,3 +5,10 @@ does> @ execute ;
 : is state @ if
 postpone ['] postpone defer!
 else ' defer! then ; immediate
+: hide ( word -- )
```
jkotlinski, line 8: I have not been doing it proper myself but maybe `: hide ("name" -- )` would clearer convey the usage. Also, this word should probably be added to the manual.

jkotlinski (changes requested):

CHANGELOG.md
jkotlinski: (FORGET would be a better name, but it does not work, since it would be confused with the classic Forth-83 FORGET)

jkotlinski (approved):
Approved 👍  thank you!

PR #287, opened 2020-10-31
Hid private words of modules

burnsauce:
Here are 15 words that are undocumented and not used outside of their own source file.


forth_src/defer.fs
```diff
@@ -12,3 +12,5 @@ dup latest @ - ( nt size )
 latest @ swap over +  ( srca dsta )
 dup latest !
 r> move then ;
+hide >body
```
burnsauce, line 15: Oops, this is used in `testcore`.  It could be redefined in compat and `>body` factored away here.

jkotlinski: I'll review this once everything is resolved :-)

burnsauce:
I'll push a new commit once #292 merges.  Wouldn't make sense to push it yet.

Issue #274, opened 2020-10-16
Error in sprite file : sp-x! error with column > 255


vinej:
There are bugs for  sp-x!. 
sp-x!  : extra bit is not will implemented. Was not working > 255 column
sp-x@ : new, to get the x position
sp-y@ : new  to get the y position

here is the modifications needed, but I don't know well forth, maybe better way to do it.

here
$80 c, $40 c, $20 c, $10 c,
8 c, 4 c, 2 c, 1 c,
: 80lsr [ swap ] literal + c@ ;

: 7s- 7 swap - ;

: getbit ( n addr -- )
swap 80lsr swap c@ and ;

: setbit ( n addr -- )
swap 80lsr over c@ or swap c! ;

: clrbit ( n addr -- )
swap 80lsr invert over c@ and swap c! ;

: sp-x@ ( n -- )
dup 7s- $d010 getbit \ get bit 255+
if 256 else 0 then   \ if set 256
swap    ( u n )
2* $d000 + c@ +
;

: sp-y@ ( n -- )
2* $d001 + c@ ;

: sp-x! ( x n -- )
2dup    ( x n x n )
2* $d000 + c! \ lsb
swap $100 and
if 7s- $d010 setbit
else 7s- $d010 clrbit then ;

: sp-y! ( y n -- ) 2* $d001 + c! ;

: sp-xy! ( x y n -- )
tuck sp-y! sp-x! ;

( expand width/height )
: sp-1w ( n -- ) 7s- $d01d clrbit ;
: sp-2w ( n -- ) 7s- $d01d setbit ;
: sp-1h ( n -- ) 7s- $d017 clrbit ;
: sp-2h ( n -- ) 7s- $d017 setbit ;

: sp-on ( n -- ) 7s- $d015 setbit ;
: sp-off ( n -- ) 7s- $d015 clrbit ;

: sp-col! ( c n -- ) $d027 + c! ;

( read sprite byte )
: ks
2* source drop >in @ + c@
1 >in +! '.' <> 1 and or ;
: rdb ( addr -- addr )
0 ks ks ks ks ks ks ks ks
over c! 1+ ;

( read sprite to address )
: sp-data ( addr -- )
#21 0 do refill rdb rdb rdb loop drop ;

burnsauce:
Here's a minor refactor, but I'm also new to forth.
```
: sp-x@ ( n -- x )
dup 2* $d000 + c@ swap
7s- $d010 getbit \ get bit 255+
[...]

PR #290, opened 2020-11-01
Fixed bug in SP-X! (Thanks @vinej!)

burnsauce:
Fixes issue #274


jkotlinski (approved):
I haven't tested this, but if it works, excellent :-)

PR #291, opened 2020-11-01
Changed LATEST from VARIABLE to VALUE

burnsauce:
Rationale: `latest` is almost always read, and mostly written to in ASM where there is no write overhead. Aligns with `here` being a `value`.

Saves 17 bytes of code space on the default build.


jkotlinski (approved):
makes sense!

PR #292, opened 2020-11-02
DEFINE assigns new behaviour to a word

burnsauce:
Used as an alternate to `IS`/`DEFER` with no associated overhead, `DEFINE`/`HEADER` allows a word to be declared before its definition.  Unlike `DEFER`, the word must not be used until `DEFINE`d.


jkotlinski (approved):
Neat!

PR #299, opened 2020-11-13
BYE now calls C64 warm reset vector

burnsauce:
Tested, works.


burnsauce:
Something is still wrong.

Whammo:
jmp($fffc)

PR #300, opened 2020-11-13
BYE now calls C64 warm reset vector

burnsauce:
Thanks @Whammo


jkotlinski:
I see why this change is convenient, however it greatly limits the usefulness of `BYE`.
With the reset, it will not be possible to use `BYE` to return control to some program that did a `jsr` to Durexforth.

burnsauce:
Hadn't considered this to be a possibility.

Whammo:
So, to preload the stack to the reset vector, shouldn't be a problem.

jkotlinski:
Maybe it’s doesn’t matter and BYE could even be removed?
I had a use for the current functionality once, but I am not sure I will
use Durexforth again anyway ;-)  so it doesn’t really matter too much for
me.

PR #301, opened 2020-12-09
gfx: PEEK should switch out Kernal automatically

rhalkyard:
`PEEK` is documented as a graphics word that can be used without any special preparation, but it assumed that the Kernal had already been switched out, causing incorrect results unless the Kernal had already been switched out before it was invoked.

For example:
`$10 clrcol 3 3 plot 3 3 peek .` should output `16`, but instead we get `0`

The only use of `PEEK` that I could find was in `PAINT` inside `gfx.fs` itself; I have updated `PAINT` to work with the fixed `PEEK`.

Alternatively, we could simply document that `PEEK` must be used in conjunction with  `KERNAL-OUT` and `KERNAL-IN`, but changing `PEEK`'s behavior seems more user-friendly :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  - V long line
  - Documented SEE concatenating subsequent :NONAME
  - SP-X! had bug in most significant bit
+ - GFX: PEEK fetched bitmap pixels from ROM instead of RAM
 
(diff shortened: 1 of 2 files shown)


jkotlinski (approved):
Makes sense! Really strange that I did it like that from the beginning. Thank you for the contribution!

If you have the time, you could also update CHANGELOG.md so that the fix is mentioned.

rhalkyard:
Added a CHANGELOG.md entry for the fix. Thanks!

PR #304, opened 2020-12-11
Fixes #302 and #303

### LINKED ISSUE #302: SAVE-PACK leaves environment unusable afterwards
`SAVE-PACK` does not undo its 'packing' operation after it has finished saving, leaving only 32 bytes of free dictionary space. To continue working, the user must reset and run from the saved environment or run `$9fff TOP!` to move the wordlist back to where it belongs (which is what `RESTORE-FORTH` does when starting a packed environment).

As a new user, this behavior seems a bit surprising - I can save the environment with `SAVE-FORTH` and then continue working in it, and would intuitively expect `SAVE-PACK` to work the same way.

If this behavior is intentional, it should probably be documented, but if possible, I think it would be more user-friendly for `SAVE-PACK` to leave the system in the state that it found it.

### LINKED ISSUE #303: SAVE-PRG places LATEST after turnkey module
`save-prg` is intended to allow the programmer to use `#32 here +` as a base address for free usage.

However, the current implementation places a 0 as a `latest` anchor in the dictionary at the time of the `save-prg` call.  This will obviously be after the `turnkey` words, which are removed from the dictionary after `restore-forth` is called at startup, whereupon `here` rewinds to well before our anchor.

burnsauce:
Adjustments to `turnkey` words to produce a better user experience.

Fixes #302 by restoring forth after `save-pack`
Fixes #303 by locating `latest` at a constant 0 byte
Clarifies some documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  - V long line
  - Documented SEE concatenating subsequent :NONAME
  - SP-X! had bug in most significant bit
+ - RESTORE-FORTH did not restore start vector
+ - SAVE-PACK did not leave the system usable
+ - SAVE-PRG now points LATEST to a constant 0
 
 ## [2.0.0] - 2020-03-22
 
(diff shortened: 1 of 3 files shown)


jkotlinski:
Thank you! I removed the CHANGELOG additions, as I supposed there is no relevant change compared to latest release (2.0.0). Maybe it's time for a new release soon.

Issue #95, opened 2015-03-14
C128 version?


GoogleCodeExporter:
```
80 columns would be nice..?
```

Original issue reported on code.google.com by `kotlin...@gmail.com` on 27 Oct 2012 at 11:59

jkotlinski:
Seems to be one at http://forth128.blogspot.se/

polluks:
Pretty cool! Of course there are some changes since August. Is there any download?

rhalkyard:
I had the same idea, and got it sort-of working fairly without too many changes; I've put a work-in-progress branch up at https://github.com/rhalkyard/durexforth/tree/c128-port.

So far, both 80 column and 40 column text modes work (including switching between them with escape-X), and it can load and compile source and operate interactively. I've added `fast` and `slow` words that work like their BASIC 7.0 counterparts, and a `fast?` word to return the current fast/slow state. `v` and graphics do not work yet, since they have some C64-specific assumptions in them, but those should be fairly easy to overcome.

Eventually it would be nice to support both the C64 and 128 in a single codebase. Would it be reasonable to add a built-in word to determine the platform that we are running on, so that the same code can do the right thing on both systems?

I also have a Plus/4 sitting around. If I get bored this winter I might try porting to that too ;)

jkotlinski:
Hi!
Wow, that's very exciting and impressive!
I had a glance, and at least so far, it looks like not many changes were needed?
If it's not more, it would be enough to add an Acme define (if there is none already) to conditionally apply the changes.
For $801/$1c01, I think it would be appropriate to add a CONSTANT for that.
If more changes are needed, certainly it would not hurt to add a word that identifies the target.

rhalkyard:
Thank you! I'm impressed with how easy it was to get it working - it's a very easy-to-understand codebase.

`v.fs` and `gfx.fs` will need to handle RAM/ROM switching differently on the 128, and it might be nice to selectively call `fast` in a few places (e.g. when compiling), so it probably would be worthwhile to have a word. Any objections to calling it `c128?`?

jkotlinski:
I'm not fully aware what fast/slow does? I mean... "fast" sounds nicer than "slow" but I suspect there are circumstances when it cannot be used? :-)

polluks:
It also works in C64 mode, how about "?vdc"?

rhalkyard:
Pushed a few more changes that got `ls` and `v` working (though `v` only works correctly in 40 column mode).

[...]

PR #305, opened 2020-12-15
Commodore 128 Port

rhalkyard:
After the discussion in #95, I reorganized the Makefile so that conditional compilation (`[if]`-`[else]`-`[then]`) was no longer needed.

Now, everything in `forth_src/` is common to both systems, and `forth_src/c64` and `forth_src/c128` contain code that is specific to those machines. `v` and `gfx` now include `vcore` and `gfxcore`, which contain their platform-specific parts.

There are also a couple of new source files:

- `kernal` is intended to define platform-specific constants for kernal variables and subroutines; right now this is only `k-tblx`, the TBLX zeropage location that indicates the current cursor line (used by `more`), but a lot of constants in `v` could be factored into it for clarity.

- `far` is only for the C128, and defines 'far' versions of `!`, `@`, `c!` and `c@` that can operate on memory in other banks. `sys` on the C128 also defines a `sysfar` word that works like `sys` but can call routines in other banks.

- `platform` detects the type of system it is being compiled on, and defines a `platform` constant as either 64 or 128, depending on the system - with conditional compilation gone, there was no need for the `native-c128` compile-time constant, but I thought that platform detection might still be useful to somebody.

I have added a new section to the manual describing the 128 port and 128-specific words. I wasn't quite sure where in the table of contents it should go; feel free to move it as needed.


jkotlinski:
Thank you! How exciting! I will have a careful look in time and try to provide meaningful feedback. Looking forward!

rhalkyard:
Take your time - I realise there's a lot of changes in there. I'm new to Forth, so please forgive any novice mistakes!

forth_src/platform.fs
```diff
@@ -0,0 +1,14 @@
+marker ---get-platform---
+
+\ use IRQ vector to identify which
+\ system we are running on
+: get-platform 
+$fffe @ case
+$ff48 of 64 endof
+$ff17 of 128 endof
+abort" cannot identify platform"
+endcase ;
```
jkotlinski, line 10: It seems wasteful to do this check in runtime, both in terms of memory and CPU cycles. Could it be better to introduce a constant that is set from assembly TARGET?

forth_src/c128/vcore.fs
```diff
@@ -0,0 +1,7 @@
+code rom-kernal ;code
+code ram-kernal ;code
+: v-startup 
+$d7 c@ if
+abort" v only supports 40-column mode.
+press esc-x to switch modes."
+then ;
```
jkotlinski, line 7: i suppose `if`/`then` should be removed.

forth_src/c64/kernal.fs
```diff
@@ -0,0 +1,4 @@
+\ Definitions for kernal vars and routines
+
+\ TBLX - cursor line number
+$d6 constant k-tblx
```
jkotlinski, line 4: is "k" for "kernal"? i think "tblx" would be enough.

It's a little disturbing to have a file for a single constant, since compile times increase with the number of files. Could it be solved in some other way?

jkotlinski:
Maybe this is a big enough change to warrant mention in README.md?

xlar54:
Hows the progress on this PR?  This is very interesting indeed.

jkotlinski:
I forgot about this PR, sorry about that.

jkotlinski:
Since time passed, there now is a lot of conflicts. I will continue the work of merging this in #478

PR #306, opened 2020-12-16
Fixed EOL and file mode

polluks:


docs/Makefile
```diff
@@ -1,5 +1,3 @@
-C1541 = c1541
-AS = acme
```
jkotlinski, line 2: Was this change intentional?

docs/tutorial.tex
```diff
@@ -1,194 +1,234 @@
...
-\subsection{Other}
-
-\begin{itemize}
-\item \href{https://github.com/jkotlinski/durexforth}{durexForth source code}
-\end{itemize}
+\chapter{Tutorial}
+
+\section{Interpreter}
+
+Start up durexForth. When loaded, it will greet you with a blinking yellow
```
jkotlinski, line 5: 🤔 Are long lines in this file a problem?

jkotlinski:
Is there a practical value for you to make these changes?
For me personally it is not a problem if linebreaks are not consistent between files. It seems to work fine anyway.

docs/Makefile
polluks: Keep it simple, you will never assemble in the docs.

polluks:
Linebreaks are just for convenience, end of the year clean-up ;-)

jkotlinski:
I see... well personally the line breaks do not bother me because i never
had a problem with that, as far as I could remember.
Maybe its different for others?
Generally speaking, I’m against ”change every line” kind of changes,
because they mess a bit with history and make ”git blame” a little less
convenient to work with.
But if it makes things more convenient for someone out there, I’m all for
it.

I suspect the effort to reformat .tex files to a particular line length is
futile. The TeX editors that I tried simply put each paragraph on a very
long line. The easy and permanent solution to that behavior is to just
accept it :-)

burnsauce:
I'm also against these, they really clutter up the view of the repo and the blame, etc.

docs/Makefile
jkotlinski: Ah! I didn't realize it was the docs folder :D

jkotlinski:
Merged! Better have it over with.

jkotlinski:
OK, I reverted this. The line ending changes caused merge conflicts with other PR's.
It is something that could be revisited once there are no big PR's in progress.

jkotlinski:
It might not be practical to accept the line end changes. The other changes are fine.

PR #310, opened 2020-12-17
Rework io

### LINKED ISSUE #252: Create Printer example
See discussion in #233 

This could be added to the manual, and/or example source code on disk.

rhalkyard:
Here's a proposal to close #252 and improve the flexibility of the IO words as discussed in that issue:

I have removed `OPENW` and `CLOSEW` and replaced them with two source files:

- open.fs contains 'raw' open and close routines without error handling, for use with `V`
 
- io.fs contains IO words for all Kernal routines, with error handling

The reason for splitting them up is twofold: 

- `V` only uses the un-error-checked `(OPEN)` and `(CLOSE)`, so by breaking them out into a separate file, we avoid cluttering the default dictionary and don't take up any extra cartridge space.

- The Commodore 128 port that I'm working on (see #95 and #305) needs to do some additional setup as part of `(OPEN)`, so it will become a platform-specific component anyway.

Additionally, I have added a small DOS-command and drive-status utility in dos.fs, which also serves as an example of using IO words for both input and output.

Finally, I've written docs for the new words, and added a small printer example to the tutorial.


rhalkyard:
@jkotlinski do you have any strong preference about `OPEN` and `CLOSE` staying in disk.asm? I figured that this way, people could strip them out if they didn't need them.

Let me know what you think about the `(OPEN)` vs `OPEN` naming scheme - I wanted to make it possible for programs to handle errors themselves, without forcing everybody to have to bother with that.

docs/words.tex
```diff
@@ -405,6 +405,38 @@ \section{Disk I/O}
...
+\subsection{Low-Level Device I/O}
+
+For more advanced uses, words corresponding to the standard Commodore Kernal IO routines are available by including \texttt{io}.
+
+\begin{description}
+    \item[open ( filenameptr filenamelength secondary-addr file-number -- )] Open a file.
+    \item[chkin ( file-number -- )] Selects a file as the current input device.
+    \item[chkout ( file-number -- )] Selects a file as the current output device.
+    \item[clrch ( -- )] Resets input and output to the keyboard and screen.
+    \item[readst ( -- status )] Returns the status from the last IO operation.
```
jkotlinski, line 431: It would help to know what status value means.

docs/words.tex
```diff
@@ -405,6 +405,38 @@ \section{Disk I/O}
...
+
+    \item[dos command ( -- )] Sends \texttt{command} to secondary address 15 on the current device, and prints the drive's response. Note that the remainder of the line is treated as part of the command. This makes it possible to refer to file names that contain spaces, but means that \texttt{dos} and its command should be on their own line, or the last words on a line. Example: \texttt{dos scratch0:old file} will delete a file named ``\texttt{old file}''.
+\end{description}
+
+\subsection{Low-Level Device I/O}
+
+For more advanced uses, words corresponding to the standard Commodore Kernal IO routines are available by including \texttt{io}.
+
+\begin{description}
+    \item[open ( filenameptr filenamelength secondary-addr file-number -- )] Open a file.
```
jkotlinski, line 427: To me it would seem more natural to give file# first, secondary address later. Because from BASIC it's e.g. OPEN "FOO",8,1

jkotlinski (commented):
Added a lot of random comments. Thanks for the awesome contribution, I'm impressed!

forth_src/dos.fs
```diff
@@ -0,0 +1,14 @@
+require io
+
```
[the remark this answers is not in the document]
rhalkyard: Probably even safer to just use $ hex values.

forth_src/io.fs
```diff
@@ -0,0 +1,75 @@
...
+: chkin ( file# -- )
+dup (chkin) ioerr ;
+
+\ Set output device to open file #
+\ 'easy' version, aborts on error
+: chkout 
+dup (chkout) ioerr ;
+
+\ accept only works for keyboard input
+: faccept ( addr len -- len )
```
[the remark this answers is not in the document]
rhalkyard: Fair enough; I was originally going to use it for reading drive status, but I realise there are better ways to do that now :). I'll remove it.

forth_src/io.fs
```diff
@@ -0,0 +1,75 @@
...
+\ Get a byte from input device
+code chrin ( -- chr )
+dex, w stx, 0 lda,# msb sta,x
+$ffcf jsr, \ CHRIN
+w ldx, lsb sta,x
+;code
+
+\ handle returned result value
+: ioerr ( file# result -- )
+if clrch close rvs ." io err" cr abort
```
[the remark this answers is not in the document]
rhalkyard: That `ioresult` idea is great; definitely makes things a bit cleaner.

Reading the error channel would definitely be more user-friendly, though correctly handling cases like a nonexistent device, or devices that aren't disk drives, will add a bit more complexity. My intent with `ioerr` was to just be a minimal safety net, but if you think the extra complexity is warranted, I'll add it in. I suppose we could expose `_errorchread` from disk.asm as a word.

forth_src/dos.fs
```diff
@@ -0,0 +1,14 @@
...
+\ send command string to drive and
+\ print response
+: send-cmd ( addr len -- )
+0 0 15 15 open 
+15 chkout type \ send command
+clrch 15 chkin
+refill source type cr \ print result
+clrch refill 15 close ;
```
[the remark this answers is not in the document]
rhalkyard: The first `refill` reads the drive status into the input buffer, `source type` prints it but doesn't consume the buffer, so the second `refill` is needed in order to flush the buffer and prevent it being interpreted as input.

 It's kind of a hack, yes, but it seemed to be the quickest and easiest way to do it with the existing wordset.

docs/tutorial.tex
```diff
@@ -63,6 +63,27 @@ \section{Console I/O Example}
...
+include io
+
+: print-hello
+4 device ( use device #4 )
+0 0 7 47 open ( open address 7 on device 4 as file 47 )
+47 chkout ( redirect output to file 47 )
+." Hello, printer!" cr
+clrch ( stop input and output redirection )
+." Hello, screen!" cr
+47 close ( close file 47 ) ;
```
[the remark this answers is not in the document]
jkotlinski: Nevermind this comment! Greg convinced me first CLRCHN then CLOSE is the way to do it. I have been doing it wrong all these years!

forth_src/dos.fs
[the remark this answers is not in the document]
jkotlinski: Right! or # for decimal

forth_src/dos.fs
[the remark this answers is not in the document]
jkotlinski: Ah! That seems like the right way to do it!

forth_src/io.fs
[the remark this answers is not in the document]
jkotlinski: I didn't think about all those special cases. I think it might be nice for `dos` and `send-cmd` since those are specifically intended for use with drives. Not for `open`/`close` which are supposed to work with anything.

forth_src/dos.fs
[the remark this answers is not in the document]
jkotlinski: Thinking about it further, it doesn't really work. Imagine following case:

    : fmt s" N0:NAME,ID" send-cmd ;
    fmt fmt fmt fmt fmt

This would only format the drive once... right?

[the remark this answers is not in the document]
jkotlinski: I came to think of a nice can of worms now. What if this code runs while a file is being INCLUDED?

Whammo:
If DEVICE were to fall into _errorchread in disk.asm, it could indicate whether the choice is valid or not.

rhalkyard:
> 
> 
> If DEVICE were to fall into _errorchread in disk.asm, it could indicate whether the choice is valid or not.

Unfortunately, trying to read the error channel of a nonexistent device causes the kernal to go into a deadlock, so it isn't quite that straightforward: https://codebase64.org/doku.php?id=base:reading_the_error_channel_of_a_disk_drive

It is possible to work around that issue, but a check like that wouldn't work with devices like printers and plotters, and I'm cautious about adding too much special-case logic to the IO words - ideally I think they should map very closely to the underlying Kernal calls.

Whammo:
Very true! Thanks!  feature creep <> lean

forth_src/dos.fs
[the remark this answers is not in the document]
rhalkyard: > 
> 
> I came to think of a nice can of worms now. What if this code runs while a file is being INCLUDED? I suspect the new words would not restore things nicely. Same with the already existing words in disk.asm (except INCLUDED itself).
> 
> Food for thought :D No action required.

Yeah, there's definitely potential for Bad Things to happen if immediate-mode code tries to use a file# that INCLUDE already has open, or doesn't clean up after itself, but I think that it should be safe if used carefully.

forth_src/io.fs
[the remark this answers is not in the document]
rhalkyard: `dos` and `send-cmd` already prints the error channel after sending a command (and can be used to just read the error channel if you send an empty command), I agree with you about `open`.

I did find out that the kernal routines for `open`, `chkin` and `chkout` leave an error number in A when they fail, so I've reworked those words to capture that error number, and given `ioerr` some more meaningful error messages.

rhalkyard:
> Very true! Thanks! feature creep <> lean

I like the potential of building up a more 'friendly' IO library on top of this base, though - one that programs could include if they wanted and leave out if they didn't.

I'm not sure how well they'd map to Forth, but having one-word equivalents to BASIC's `INPUT#`, `GET#` and `PRINT#` statements might be nice.

Also, the ability to do `gfx` and `turtle` graphics output to printers and plotters! I'm working on some ideas for that...

Whammo:
Or at least these equates:
 
$ffb7 constant readst
$ffba constant setlfs
$ffbd constant setnam
$ffc0 constant open
$ffc3 constant close
$ffc6 constant chkin
$ffc9 constant chkout
$ffcc constant clrchn
$ffcf constant chrin
$ffd2 constant chrout

rhalkyard:
> 
> 
> Or at least these equates:

Yeah, I've throught about making an `include`-able file with the addresses of useful Kernal routines and variables - especially because on the C128, some of them are in different places, so referring to them by name rather than by address makes it easier to write code that runs on both systems.

forth_src/io.fs
```diff
@@ -0,0 +1,79 @@
...
+
+\ Set input device to open file #
+\ 'easy' version, aborts on error
+: chkin ( file# -- ) 
+chkin? ioerr ;
+
+\ Set output device to open file #
+\ 'easy' version, aborts on error
+: chkout ( file# -- )
+chkout? ioerr ;
```
jkotlinski, line 79: I think everything looks kind of good, the one thing I'm not sure about are these "easy" versions. Maybe it would make things in practice simpler, to only have one set of open/chkin/chkout words that always return a status?
When looking in C and Forth libraries, this to me seems like the normal thing to do.

I think cc65 is a good reference: https://cc65.github.io/doc/funcref.html#ss2.11

For those who would like an "easy" version without error handling, it could always be added later?
What do you think?

jkotlinski: As a contrary reference, 64Forth OPEN/CHKIN/CHKOUT seem to have in-built error handling - or at least those functions do not return any value to indicate success or failure.

Forth is kind of difficult because the only thing you know, in ease of use and performance it is "supposed" to be somewhere between BASIC and C - and that leaves quite a lot of space for tradeoffs and design choices.

jkotlinski: I think what definitely speaks for having `ioerr` around is that the printer example now looks simple and good.

Changing printer example to e.g. `0 0 7 47 open ioerr ( open address 7 on device 4 as file 47 )` would make it a bit less nice, but it's still OK.

Leaving all error handling to be implemented by the user, then it feels like the printer example will not be nice anymore.

I would consider renaming `ioerr` to `ioabort`. I think it becomes a bit easier to understand what it does in that way.

forth_src/open.fs
```diff
@@ -0,0 +1,35 @@
+\ Open a file, no error handling
+\ returns 0 on success, file # on error
```
jkotlinski, line 2: This comment looks wrong, should it be "\result=0 on success, or OPEN error code" ?

forth_src/open.fs
```diff
@@ -0,0 +1,35 @@
...
+0 lda,# \ A is only valid on error
+:+
+w ldx,
+inx, inx,
+lsb sta,x
+lsb 1- lda,x lsb 1+ sta,x
+0 lda,# msb sta,x msb 1+ sta,x
+;code
+
+\ Close a file
```
jkotlinski, line 29: For ultimate correctness I guess it should be "close logical file"

forth_src/io.fs
```diff
@@ -0,0 +1,79 @@
+require open
+
+\ Set input device to open file #
+\ returns 0 on success, file # on error
```
jkotlinski, line 4: This comment looks wrong, should it be "\result=0 on success, or CHKIN error code" ?

forth_src/io.fs
```diff
@@ -0,0 +1,79 @@
...
+0 lda,# \ A is only valid on error
+:+
+w ldx, dex, 
+lsb sta,x
+0 lda,# msb sta,x
+;code
+
+\ Set output device to open file #
+\ returns 0 on success, file # on error
+code chkout? ( file# -- file# result )
```
jkotlinski, line 19: This comment looks wrong, should it be "\ result=0 on success, or OPEN error code" ?

forth_src/io.fs
jkotlinski: I am sorry for all the "loud thinking" but my current opinion is, as a summary:

 - drop the "easy" versions, only keep one set of `open`/`chkin`/`chout`
 - rename `ioerr` to `ioabort` and document it
 - in documentation, rename "result" of above word signatures to "ioresult" to clarify that these words are supposed to be used together

forth_src/viceutil.fs
```diff
@@ -54,6 +56,6 @@ else emit then loop
 $a emit 1 ;
 
 : dump-labels base @ >r hex
-s" words" 1 openw
+s" words,w" 1 1 open 1 chkout
```
jkotlinski, line 59: I think this would be an excellent use case for the new `ioerr`/`ioabort`!

CHANGELOG.md
```diff
@@ -29,12 +29,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+ - Utility for drive status and DOS commands in DOS:
 ### Fixed
  - SP-X! had bug in most significant bit
  - GFX: PEEK fetched bitmap pixels from ROM instead of RAM
  - V did not compile in DECIMAL mode.
  - Documented SEE concatenating subsequent :NONAME
- - Made QUIT do CLRCHN instead of CHKIN.
+### Removed
+ - Removed undocumented OPENW and CLOSEW words.
+ - Made QUIT do CLRCHN instead of CHKIN
```
jkotlinski, line 41: This should be moved up to `### Fixed` section.

jkotlinski:
> I like the potential of building up a more 'friendly' IO library on top of this base, though. I'm imagining a library that programs could include if they wanted and leave out if they didn't.
> 
> I'm not sure how well it would map to Forth, but it would be nice to have some one-word equivalents to BASIC's `INPUT#`, `GET#` and `PRINT#` statements so you don't have to keep juggling `chkin` and `clrchn` - that's where I'd put more thorough error checking and input validation.

Adding BASIC equivalents should be possible. It's nice for beginners, the main problem is that those words would have a lot of overhead. But as long as it is opt-in, I cannot see a problem with adding that.

forth_src/io.fs
rhalkyard: I like this idea a lot. `open ioabort` isn't *really* that much more to type than just `open`. I think it hits a nice middle ground of 'more expressive than BASIC, but easier than asm or C'.

jkotlinski (approved):
Brilliant! Thank you for the contribution. Happy holidays!

PR #313, opened 2020-12-19
Fix #309 by updating dump-labels to use dowords

### LINKED ISSUE #309: viceutil.fs is broken

rhalkyard:
Fixes #309. Also fixes a `dowords` bug that caused the lambda to be called on the invalid nametoken at the end of the dictionary.

### DIFF
--- a/forth_src/viceutil.fs
+++ b/forth_src/viceutil.fs
@@ -9,12 +9,12 @@ Then, load the file from VICE
 monitor using 'll "words"'
 )
 
-\ TODO broken until TRAVERSE-WORDLIST
-: dump-labels base @ hex
-s" words" 1 openw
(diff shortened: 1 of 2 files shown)


rhalkyard:
@burnsauce, as the official split-dictionary wizard, would you mind looking at my `dowords` changes in 5c76e8d? The original behavior looked like a bug to me, but if there's a reason it needs to work that way, I can add a check to my code instead.

burnsauce (approved):
Looks sane, although you might want to spin this out to a separate PR for clarity.

Don't know how I wrote such an obvious bug, but good catch!

jkotlinski:
Awesome! Thank you for the contribution!

PR #340, opened 2021-04-22
Update interpreter.asm

Whammo:
This avoids rom cart init- passes through non-CIA NMI's (Restore) silently, and executes quit at run/stop Restore.
Replaces restore_handler entirely in interpreter.asm  YaY !

### DIFF
--- a/interpreter.asm
+++ b/interpreter.asm
@@ -43,16 +43,13 @@ stop_restore
 	cmp	#$7f		; compare with [stp] down
 	        		; if not [stp] or not just [stp] exit
     bne	kernal_nmi	; if not [stop] restore registers and exit interrupt
-    pla
-    pla
-    tax
-    jmp QUIT
+    jsr $f333       ; set screen, keyboard, devices untalk, unlisten
 
 brk_handler
    pla
    pla
    tax              ; restore xr for QUIT
-   jmp skip_clall
+   jmp QUIT
    
 quit_reset  ; execute once at start
     sei
@@ -118,10 +115,6 @@ keep_nmi
 
     +BACKLINK "quit", 4
 QUIT
-    jsr $ffe7       ; CLALL. Clear file table; call CLRCHN.
-    
-skip_clall
-
     jsr keep_nmi
 
     ; resets the return stack
--- a/interpreter.asm
+++ b/interpreter.asm
@@ -107,7 +107,7 @@ keep_nmi
     stx     SOURCE_ID_MSB
     stx     SAVE_INPUT_STACK_DEPTH
     stx     READ_EOF
-    jsr $ffe7       ; CLALL. Clear file table; call CLRCHN.
+    jsr     $ffcc   ; CLRCHN
     pla
     tax
     rts
--- a/interpreter.asm
+++ b/interpreter.asm
@@ -40,6 +40,8 @@ stop_restore
    jsr	$f6bc		; increment real time clock
    jsr	$ffe1		; scan stop key 
    bne	kernal_nmi	; if not [stop] restore registers and exit interrupt
+   
+brk_handler
    pla
    pla
    tax              ; restore xr for QUIT
@@ -51,6 +53,11 @@ quit_reset  ; execute once at start
     sta $318
     lda #>restore_handler
     sta $319
+
+    lda #<brk_handler
+    sta $316
+    lda #>brk_handler
+    sta $317
     
 keep_nmi
     cli ; still have to
--- a/interpreter.asm
+++ b/interpreter.asm
@@ -43,13 +43,16 @@ stop_restore
 	cmp	#$7f		; compare with [stp] down
 	        		; if not [stp] or not just [stp] exit
     bne	kernal_nmi	; if not [stop] restore registers and exit interrupt
-    jsr $f333       ; set screen, keyboard, devices untalk, unlisten
+    pla
+    pla
+    tax
+    jmp QUIT
 
 brk_handler
    pla
    pla
    tax              ; restore xr for QUIT
-   jmp QUIT
+   jmp skip_clall
    
 quit_reset  ; execute once at start
     sei
@@ -115,6 +118,10 @@ keep_nmi
 
     +BACKLINK "quit", 4
 QUIT
+    jsr $ffe7       ; CLALL. Clear file table; call CLRCHN.
+    
+skip_clall
+
     jsr keep_nmi
 
     ; resets the return stack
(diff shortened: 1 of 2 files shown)


Whammo:
Set NMI vector only once at startup, quit_reset for compatibility. (IDE64?)

Whammo:
I want to do something snazzy with the brk vector. Ideas? ( breakpoint instead of breaking durexforth )

Whammo:
We don't need cli, QUIT txs 0 to sr

Whammo:
Noes, we must cli in QUIT because txs is stack pointer not status

Whammo:
brk instruction no longer breaks Durexforth

jkotlinski:
Cool, do I understand right, this fixes RS-232? Is it in a good state to merge?

Whammo:
It has not been tested in the current state. I've tested it as a patch on the release.

Whammo:
I'll have a better answer for you soon.

Whammo:
Roger, Houston. We are go for TLI.

Whammo:
no wait!!!

Whammo:
Checking for data stack effects

Whammo:
Yeah, sorry. Good to go!

Whammo:
Doggone it- I'm pretty sure run/stop restore causes data stack issues, that won't do. it's probably:
 
  jsr	$ffe1		; scan stop key 

let me take a look at it.

Whammo:
Okay the change I made didn't change the behavior, and I'm seeing similar stuff in the build without this.
 I think I'm good.

Whammo:
I had to pull that back, because it needs to set i/o defaults.

jkotlinski:
Is it good to go? Sorry that I have to ask. I find it difficult to understand, because the nature of the problem :-)

Whammo:
RS232 is working,  regular operation is stable I think it's good to go.

Whammo:
Just to be safe, I'm going to revert what I reverted last- it did remediate the stack issues, Thanks cminter!

Whammo:
Also, the stack issues I mentioned earlier are not issues. The deeper the code goes, the deeper the stack grows.
  It surprised me, that there was so much on the stack.

interpreter.asm
```diff
@@ -43,7 +43,7 @@ stop_restore
 	cmp	#$7f		; compare with [stp] down
 	        		; if not [stp] or not just [stp] exit
     bne	kernal_nmi	; if not [stop] restore registers and exit interrupt
-
+    jsr $f333       ; set screen, keyboard, devices untalk, unlisten
```
jkotlinski, line 46: wouldn't this be taken care of by QUIT call to CLRCHN?

Whammo:
CLRCHN leaves files open. Run/stop Restore during LS will show this if you run LS again after Run/stop restore. I kept $ff33 from running after brk for reentrant purposes.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

jkotlinski:
I think this looks suspicious because ff33 is the default CLRCHN
implementation on C64. Why does it make a difference to call it twice?

jkotlinski:
...sorry, i mean f333 not ff33 but question remains :-)

Whammo:
I will have to take a look at it and get back to you.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

Whammo:
I have a better idea.

Whammo:
Crud. CLRCHN affects x

interpreter.asm
```diff
@@ -76,14 +107,14 @@ quit_reset
     stx     SOURCE_ID_MSB
     stx     SAVE_INPUT_STACK_DEPTH
     stx     READ_EOF
-    jsr     $ffcc   ; CLRCHN
+    jsr $ffe7       ; CLALL. Clear file table; call CLRCHN.
```
jkotlinski, line 110: This use of CLALL might cause problems, as it does not CLOSE any open files.

Whammo: Okay, CLRCHN it is.  Thanks for having my back on this.  :)

jkotlinski:
I think it would make sense, if possible, to have something like C128
CLOSEALL there in the future. I should have added that as a comment.

Whammo:
Even CLOSEALL only works on the specified device. Anyway, I can see instances where one wouldn't want files closed.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

interpreter.asm
```diff
@@ -83,7 +114,7 @@ quit_reset
 
     +BACKLINK "quit", 4
 QUIT
-    jsr quit_reset
+    jsr keep_nmi
```
jkotlinski, line 117: What is the purpose of not resetting $316/$318 at QUIT?

I don't mind this change but also curious why it was done :-)

interpreter.asm
```diff
@@ -23,16 +23,47 @@
...
+                    ; scan stop key 
+   	lda	$91 		; read the stop key column
+	cmp	#$7f		; compare with [stp] down
+	        		; if not [stp] or not just [stp] exit
+    bne	kernal_nmi	; if not [stop] restore registers and exit interrupt
+    
+brk_handler
+   pla
+   pla
+   tax              ; restore xr for QUIT
```
jkotlinski, line 50: Does this pla/pla/tax properly restore x even for BRK command? How does that work?

jkotlinski: ...looking at how BRK works, I guess x would be overwritten with some seemingly random value?
If so, I think it could be non-controversial to not touch $316 at all. It saves a few bytes and it does not really take away any possibilities.

interpreter.asm
Whammo: I was under the impression that QUIT was called as a matter of course with ';' I have since resolved that Idea, and I really can't see any reason this still needs to be done.

interpreter.asm
Whammo: Y is at the top of the return stack at this point, and X is underneath. QUIT then resets the return stack preserving X.
  It was never certain that X would be valid, if Restore was initiated during IRQ or while assembly has X stored away, but it usually was good.

jkotlinski: OK cool! Yes I haven't really run this code. So if you say it works, it's fine for me.

jkotlinski:
Merged now, future changes and bugfixes are welcome like always. Thanks for your contribution!

interpreter.asm
Whammo: > ...looking at how BRK works, I guess x would be overwritten with some seemingly random value?
> If so, I think it could be non-controversial to not touch $316 at all. It saves a few bytes and it does not really take away any possibilities.

 BRK arrives at the vector with the registers intact, return address and status on the return stack. 
  So if you insert a [ brk, } into your forth code, it'll preserve the data stack. I think this is useful.
  Also, making run/stop restore and BRK equivalent emulates the kernals' handler, and makes BRK safe.

PR #344, opened 2021-04-25
enable Makefile handling of VICE 3.4+ which has x64sc without x64

cminter:
+ using `make VICE_X64SC=1 ...` will select `x64sc` and also use the
  renamed flag for disabling 'confirm on exit' +confirmonexit

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,15 @@
 C1541   = c1541
 AS = acme
+X64 = x64
+
+X64_OPTS = -warp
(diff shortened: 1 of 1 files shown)


cminter:
The VICE package on Fedora distro only has `x64sc` and the option `confirmexit` changed to `confirmonexit` apparently at VICE 3.4.  Commit doesn't affect default build behavior.

jkotlinski (approved):

PR #345, opened 2021-04-26
include Git hash in d64 ; calculate TAG variables once

cminter:
+ include Git hash in d64 after version string (for traceability)
+ use lower case 'm' in d64 (was 'M') for warning that repo files were modified
  as d64 was generated (looks better when listing disk in upper case mode)
+ calculate TAG variables only once using shell command
  (previous form was just text in a variable that was not executed
  until later during each command which obfuscates the flow)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,10 @@
 C1541   = c1541
 AS = acme
-TAG = `git describe --tags --abbrev=0 || svnversion --no-newline`
-TAG_DEPLOY_DOT = `git describe --tags --abbrev=0 --dirty=-M`
-TAG_DEPLOY = `git describe --tags --abbrev=0 --dirty=_M | tr _. -_`
+
+TAG := $(shell git describe --tags --abbrev=0 || svnversion --no-newline)
+TAG_DEPLOY_DOT := $(shell git describe --tags --abbrev=0 --dirty=-m)
+TAG_DEPLOY := $(shell git describe --tags --abbrev=0 --dirty=_M | tr _. -_)
+GIT_HASH := $(shell git rev-parse --short HEAD)
 
 SRC_DIR = forth_src
 SRC_NAMES = base debug v asm gfx gfxdemo rnd sin ls turtle fractals \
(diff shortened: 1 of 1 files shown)


cminter:
Below is what the directory looks like.  E.g. Git hash `36e3593`

```
0 "durexforth      " DF 2a
154   "durexforth"        prg 
1     "=-=-=-=-=-=-=-="   seq 
1     "v2.0.0-m"          seq 
1     "  36e3593"         seq 
1     "=-------------="   seq 
19    "base"              prg 
8     "debug"             prg 
...
1     "=-=---=-=---=-="   seq 
83 blocks free.
```

polluks:
If prefer cc1541 because of  DEL entries instead of SEQ.

cminter:
> If prefer cc1541 because of DEL entries instead of SEQ.

Good point.  I agree that would be better.   Any ideas for making it happen with c1541?

cminter:
The second commit I pushed (3d9bf96ca2d67db27546f4f4fcafaa6657b623ad) makes the version look like below.  Notice inserted `.65` and `-m` since I had modified again after the commit.
```
0 "durexforth      " DF 2a
154   "durexforth"        prg 
1     "=-=-=-=-=-=-=-="   seq 
1     "v2.0.0.65-m"       seq 
1     "  3d9bf96"         seq 
1     "=-------------="   seq 
19    "base"              prg 
8     "debug"             prg 
...
1     "=-=---=-=---=-="   seq 
83 blocks free.

```

jkotlinski:
Looks like there are merge conflicts now. Is it good to go otherwise?

cminter:
> Looks like there are merge conflicts now. Is it good to go otherwise?

I see.  I fixed the merge confusion from Github.  I will pull and test.

cminter:
> > Looks like there are merge conflicts now. Is it good to go otherwise?
> 
> I see. I fixed the merge confusion from Github. I will pull and test.

It is good to go.

PR #359, opened 2021-08-28
Update io.fs

Whammo:
Throwing it out there.  Works from 0-30

### DIFF
--- a/forth_src/io.fs
+++ b/forth_src/io.fs
@@ -56,15 +56,10 @@ w ldx, lsb sta,x
 \ nonzero, print error message and
 \ abort.
 : ioabort ( file# ioresult -- )
-?dup if rvs case
-1 of ." too many files" endof
-2 of ." file# in use" endof
-3 of ." file not open" endof
-4 of ." file not found" endof
-5 of ." device not present" endof
-6 of ." not input file" endof
-7 of ." not output file" endof
-8 of ." missing filename" endof
-9 of ." illegal device #" endof
-." io err " endcase
+?dup if
+rvs 55 1 c! 1-
+1 lshift $a328 + @
+begin dup c@ dup 128 and 0= while
+emit 1+ repeat 128 - emit
+54 1 c!
 cr abort else drop then ;
--- a/forth_src/io.fs
+++ b/forth_src/io.fs
@@ -60,6 +60,7 @@ rvs 55 1 c! 1-
 begin dup c@ dup 128 and 0= while
 emit 1+ repeat 128 - emit
 cr abort then ;
+
 \ handle out of range ioresult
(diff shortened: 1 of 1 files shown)


jkotlinski (commented):
That's real nice thinking, however I guess it will not work when running from cartridge?

Whammo:
That, I did not consider. Isn't Cartridge at $8000?

Whammo:
Works in VICE  w/ cart image.

jkotlinski:
Oh, sorry! I just misremembered from float.fs - but the problem is only in $8000-$9fff range!

forth_src/io.fs
```diff
@@ -53,15 +53,12 @@ w ldx, lsb sta,x
 \ close, and chkin. If ioresult is
 \ nonzero, print error message and
 \ abort.
+
```
jkotlinski, line 56: was this blank line accidental, or looks better this way?

forth_src/io.fs
```diff
@@ -53,15 +53,12 @@ w ldx, lsb sta,x
...
-8 of ." missing filename" endof
-9 of ." illegal device #" endof
-." io err" endcase cr abort then ;
+?dup if
+rvs 55 1 c! 1-
+1 lshift $a328 + @
+begin dup c@ dup 128 and 0= while
+emit 1+ repeat 128 - emit
+54 1 c!
+cr abort else then ;
```
jkotlinski, line 64: the `else` looks unneeded

jkotlinski (changes requested):
brilliant idea!

forth_src/io.fs
Whammo: artifacts from removing file#

Whammo:
This takes care of the out of range and keeps all the messages available.

```forth
: berr ( ioresult -- )
?dup if
rvs 55 1 c! 1-
2* $a328 + @
begin dup c@ dup 128 and 0= while
emit 1+ repeat 128 - emit
cr abort then ;

\ handle out of range ioresult
: ioabort  ( ioresult -- ? )
dup 9 > if rev ." io err" drop
else berr then ;
```

Whammo:
Credit to Mapping the 64

jkotlinski:
excellent - thanks!

jkotlinski:
hmm, it does not seem to work for me :/

Whammo:
BERR doesn't work?

jkotlinski:
It does not compile, and after fixing compile error, e.g. ”1 IOABORT”
freezes the computer

Whammo:
ugh. Perhaps moving io.fs put it above REV

Whammo:
No, my bad- RVS instead of RSV

PR #360, opened 2021-08-28
change signature of OPEN/CHKIN/CHKOUT/IOABORT

jkotlinski:
There is no longer any reason to pass file# into IOABORT, since ABORT now closes all open files.

wdyt @Whammo ?


Whammo:
I'm on board.

Issue #55, opened 2015-03-14
cursor leaves trail at startup OK


GoogleCodeExporter:
```
boot up to see it
```

Original issue reported on code.google.com by `kotlin...@gmail.com` on 16 Jan 2009 at 4:13

PR #361, opened 2021-08-28
Update io.fs

Whammo:
It should compile now. 
oops

### DIFF
--- a/forth_src/io.fs
+++ b/forth_src/io.fs
@@ -55,7 +55,7 @@ w ldx, lsb sta,x
 \ abort.
 : berr ( ioresult -- )
 ?dup if
-rvs 55 1 c! 1-
+rvs 1 c@ 1 or 1 c! 1-
 2* $a328 + @
 begin dup c@ dup 128 and 0= while
 emit 1+ repeat 128 - emit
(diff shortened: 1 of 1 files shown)


Whammo:
Yeah, that fixes it. Sorry for the hassle.

Whammo:
you may have a different memory configuration so $01 needs to be changed Boolean?

jkotlinski:
Maybe 55 should be #55 or $55

Whammo:
I'm thinking,
```forth
 1 c@ 1 or 1 c!
 ```

jkotlinski:
Shouldn't it be `1 c@ 3 or 1 c!`? But I think `$37 1 c!` would work just as well, and take less space.

Whammo:
We only want to set bit 0., without disturbing the other bits.

jkotlinski:
Are you sure? Looking here https://sta.c64.org/cbm64mem.html , it says "%x11: BASIC ROM visible at $A000-$BFFF; KERNAL ROM visible at $E000-$FFFF."
meaning both two lowest bits need to be 1?

Whammo:
What you've added done is cool by me. I wanted to leave the >9 messages available, but it's not worth the extra code.

Whammo:
0 or 0 = 0
0 or 1 = 1
1 or 0 = 1
1 or 1 = 1
1 c@ 1 or C! Only changes bit 0 and only if it is zero



Get Outlook for Android<https://aka.ms/ghei36>
________________________________

PR #362, opened 2021-08-29
Convert IOABORT to assembly, use it in INCLUDE

jkotlinski:
@Whammo - what do you think?
I like this, because now e.g. "10 DEVICE INCLUDE FOO" will give an error message "DEVICE NOT PRESENT".
This could be very helpful.

_errorchread is still used after failed READST - I'm not sure, but right now I'd guess it's not a bad thing.

Probably some improvements should also be done for LOADB/SAVEB, I will have to think about that tomorrow. It's a bit unclear how errors should be managed for those cases. Maybe they should return an `ioresult`, or maybe something else should be done, don't know yet.

### DIFF
--- a/disk.asm
+++ b/disk.asm
@@ -272,7 +272,9 @@ INCLUDED
     jsr	SETLFS
     jsr	OPEN
     bcc	+
-    jmp ABORT
+    ldx #-1
+    sta LSB,x
+    jmp IOABORT
 +
     ldx	SOURCE_ID_LSB ; file number
     jsr	CHKIN
--- a/disk.asm
+++ b/disk.asm
@@ -272,7 +272,9 @@ INCLUDED
     jsr	SETLFS
     jsr	OPEN
     bcc	+
-    ldx #-1
+    pla
+    tax
+    dex
     sta LSB,x
     jmp IOABORT
 +
--- a/disk.asm
+++ b/disk.asm
@@ -272,9 +272,7 @@ INCLUDED
     jsr	SETLFS
     jsr	OPEN
     bcc	+
-    pla
-    tax
-    dex
+    ldx #-1
     sta LSB,x
     jmp IOABORT
 +
--- a/io.asm
+++ b/io.asm
@@ -316,6 +316,7 @@ IOABORT ; ( ioresult -- )
     jsr CHROUT
     jsr CHROUT
     jmp .cr_abort
+
 .print_basic_error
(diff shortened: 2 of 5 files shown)


Whammo:
jsr $f133 will print "i/o error #" for you and save some memory

Whammo:
Alternatively, the actual string in last byte high bit set format is at: $f0bd.

jkotlinski:
I think $f133 will only work for numbers 0-9. but i can save some bytes by reusing the string output routine!

disk.asm
```diff
@@ -272,7 +272,9 @@ INCLUDED
     jsr	SETLFS
     jsr	OPEN
     bcc	+
-    jmp ABORT
+    ldx #-1
+    sta LSB,x
```
jkotlinski, line 276: this is a bit hacky maybe, but it seems to work.

disk.asm
```diff
@@ -272,7 +272,9 @@ INCLUDED
     jsr	SETLFS
     jsr	OPEN
     bcc	+
-    jmp ABORT
+    ldx #-1
```
jkotlinski, line 275: this is a bit hacky maybe, but it seems to work.

io.asm
```diff
@@ -278,3 +288,58 @@ RESTORE_INPUT
...
+    lda LSB-1,x
+    asl
+    tax
+    lda $a326,x
+    sta W
+    lda $a327,x
+    sta W+1
+
+.print_msb_terminated_string
+    jsr CLRCHN
```
jkotlinski, line 331: i hope `close_all_logical_files` in `ABORT` will still work safely, even if I call CLRCHN first.

Whammo:
If you would be so kind as to attach your latest or next compile in reply to this, I'd like go over it. Thanks

jkotlinski:
[durexforth.d64.gz](https://github.com/jkotlinski/durexforth/files/7074700/durexforth.d64.gz)
here you go!

io.asm
Whammo: Theoretically, it isn't that difficult to see how many files and which are open. I'll look around and maybe prototype something up.

Whammo: CLRCHN just sets screen and keyboard and sends UNTALK or UNLISTEN to appropriate devices.

jkotlinski: hopefully it will work just fine still.

PR #365, opened 2021-08-31
Update disk.asm

Whammo:
Drop a couple bytes and fix SA:
```
A command to the device is sent as a secondary address on the serial
  bus after the device number is sent during the serial attention
  handshaking sequence. If no secondary address is to be sent, the Y index
  register should be set to 255.
```


Whammo:
SA 15 IS the command channel.  SA = channel. 
Pardon me.

PR #378, opened 2021-11-08
Update interpreter.asm

Whammo:
Saving the processor state before QUIT.
  makes possible: 
 : continue $14 @ sys ;

### DIFF
--- a/interpreter.asm
+++ b/interpreter.asm
@@ -46,8 +46,18 @@ stop_restore
 
 brk_handler
    pla
+   sta $30e         ; yr
    pla
    tax              ; restore xr for QUIT
+   sta $30d         ; xr
+   pla
+   sta $30c         ; ar
+   pla
+   sta $30f         ; sr
+   pla
+   sta $14          ; store rti address where Kernal SYS stores address
+   pla
+   sta $15          ; enables " : continue $14 @ sys ; "
    jmp QUIT
 
 quit_reset


Whammo:
I think I was hasty in offering this- it doesn't account for the return stack pointer that QUIT resets. 

   I don't foresee a problem with using $310 to store sp before QUIT,  
   Perhaps I also overestimate the desirability of saving and restoring the processor state?

jkotlinski:
What would be the typical use case for QUIT ... CONTINUE ?
Does CONTINUE exist in other Forths?

polluks:
I found this https://www.reddit.com/r/Forth/comments/7rcxvn/continue_and_continue_in_forth/

Whammo:
We may be in uncharted territory. I don't think Chuck Moore is available for naming. Although Elizabeth Rather chimes in frequently.

Whammo:
QUIT doesn't actually save the processor state, a hardware interrupt saves the processor state, and then calls QUIT.
 It's more of a BREAK ...RESUME

Whammo:
The idea here is that it takes a good deal of space to put another NMI routine in memory to save the processor state upon RUN-STOP RESTORE. 
  Saving and restoring the processor state, and saving and restoring system state allows task switching and thereby multi-tasking and possibly multi-user.

Whammo:
Better left for a future application.

PR #379, opened 2021-11-16
Update sys.fs

Whammo:
This is the corollary of the changes to interpreter.asm
Rather than
: continue $14 @ sys ;


Whammo:
Better left for a future application.

Issue #390, opened 2021-12-28
Documentation: "press RESTORE to quit program" correct?


Low012:
I have burned durexForth to an EPROM today and created a cartridge from it. Since I am new to Forth and durexForth, I read _durexforth-v3-0-0.pdf_ and played with the example code. The manual mentions on pages 8 and 14 that pressing the **RESTORE** key quits the currently running program.

I noticed that I had to press **RUN/STOP + RESTORE** to stop a running program instead of just **RESTORE**. Is this a mistake in the documentation or am I missing something?

Whammo:
That is true. It was changed in 3.0.0. when we upgraded NMI handling to allow rs232.
  This issue will ensure that this is resolved. Thank you.

jkotlinski:
Oh... I guess there is no way to keep the old behavior, right?

I reviewed the NMI handler. Is the "jsr $f6bc" call needed? Looks like it maybe could be removed safely? I didn't test this on real hardware yet...

`
stop_restore
    jsr $f6bc       ; increment real time clock
                    ; scan stop key
    lda $91         ; read the stop key column
    cmp #$7f        ; compare with [stp] down
                    ; if not [stp] or not just [stp] exit
    bne kernal_nmi  ; if not [stop] restore registers and exit interrupt
`

Whammo:
jsr $f6bc can be safely removed. It's there because it's in the Kernal handler.

 The issue with QUIT on NMI from RESTORE is distinguishing it from other non-CIA NMI's.
Ignoring non-CIA NMI's unless stop=true is the default Kernal behavior.

Whammo:
By default, RESTORE will start a cart if present, but we've dealt with that. This can be easily changed to call QUIT ON RESTORE.
This frees up RUN-STOP RESTORE for other purposes.
 On a real 64, you might have to hit RESTORE a couple times for it to catch the  RUN-STOP, but will have already QUIT. :D

Whammo:
$FF = no key pressed
$7F = stop key pressed

![image](https://user-images.githubusercontent.com/38296423/147713991-105ceca4-cc6b-4144-a090-83182193fea8.png)

Whammo:
This is where every NMI falls that was not initiated by the CIA.
I can make RESTORE call QUIT, and RUN STOP RESTORE call ABORT .

```asm
stop_restore
    jsr $f6bc       ; increment real time clock
                    ; scan stop key
    lda $91         ; read the stop key column
[...]

PR #391, opened 2021-12-28
Update docs concerning RUN/STOP + RESTORE

Whammo:
#390


jkotlinski:
skipping this one, the behavior was reverted

PR #392, opened 2021-12-30
 Revert to 2.0.0 RESTORE handling

Whammo:
This is as tight as I can make it.
It doesn't deal with RUN STOP RESTORE.

### DIFF
--- a/interpreter.asm
+++ b/interpreter.asm
@@ -41,7 +41,7 @@ brk_handler         ; all non-CIA NMI
     pla
     tax              ; restore xr for QUIT
     sei
-    
+quit_reset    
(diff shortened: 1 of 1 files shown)


jkotlinski:
Cool! And this is also compatible with RS-232?

Whammo:
RS232 will work, any NMI not from CIA 2 and the RESTORE key calls quit.

jkotlinski:
That is probably Fine, right?

Whammo:
It's actually excellent! There are no complications for altering the NMI vector for hardware drivers because QUIT is always there for consistent behavior.

PR #393, opened 2021-12-31
Update disk.asm

Whammo:
use logical file number table to close_all_logical_files:

### DIFF
--- a/disk.asm
+++ b/disk.asm
@@ -305,5 +305,5 @@ close_all_logical_files:
     beq +       
     lda $259,x ; logical file number table
     jsr CLOSE  ; decrements $98 
-    beq -
+    bcc -
 +   rts
--- a/disk.asm
+++ b/disk.asm
@@ -305,5 +305,5 @@ close_all_logical_files:
     beq +       
     lda $259,x ; logical file number table
     jsr CLOSE  ; decrements $98 
-    bcc -
+    jmp close_all_logical_files:
 +   rts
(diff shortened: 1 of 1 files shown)


disk.asm
```diff
@@ -301,12 +301,9 @@ INCLUDED
...
-    pla
-    tax
-    dex
-    bne -
-    rts
+-   ldx $98    ; number of open files
+    beq +       
+    lda $259,x ; logical file number table
+    jsr CLOSE  ; decrements $98 
+    bcc -
```
jkotlinski, line 308: Let's assume if CLOSE returns with carry set, it indicates an error.

Would it still make sense to close the remaining files?

I.e. replace `bcc -` with `jmp close_all_logical_files`?

jkotlinski: A small note: For C128, the logical file number table has a different location, so this code is not directly portable.
Instead, the kernal routine CLOSE_ALL could be used to close all open files.

Whammo: That's why you make the big bucks!  $$$

jkotlinski: ...wait. C128 CLOSE_ALL does not work like I thought, it works per-device. So a routine like this is still useful.

Whammo: They couldn't fit this into the Kernal.

jkotlinski:
Sorry, I considered it but I'm not picking up this one. The old code is inefficient, but it doesn't really matter since it only runs in exceptional cases. So then I value more that it's portable.

Issue #257, opened 2020-05-03
Add pad to the documentation


polluks:
Maybe to system state?

Whammo:
Or compatibility?

jkotlinski:
@burnsauce If it's not too much work, I'd welcome a pull request with your commit that fixes this issue!

PR #402, opened 2022-01-04
Documents the pad, fixing #257

burnsauce:
I have edited this but not compiled the pdf yet, so it needs testing.


jkotlinski:
Thank you! :3

Issue #288, opened 2020-10-31
Factor Kernal swapping to SYS


burnsauce:
`rom-kernal` of `gfx` and `kernal-in` of `v` are similar, as are their counterparts.

It seems like these words should be factored into `sys` or somewhere.

Whammo:
Perhaps this is a non-issue.
I can do like:
```forth
code bank
w       sta,
$01     lda,
$f8     and,#
w       ora,
$01     sta,
;code
```
And v
```forth
\ ram + io + kernal rom
code rom-kernal
$06 lda,# ' bank jsr, ;code
\ ram + io + ram
code ram-kernal
$05 lda,# ' bank jsr, ;code
```
And gfx
```forth
code kernal-in
6 lda,# ' bank jsr, cli, ;code
code kernal-out
sei, 5 lda,# ' bank jsr, ;code
```
 But it adds a couple bytes and it not something that should be done lightly.

burnsauce:
I think this issue has been resolved: it's not worth the savings to factor these commonly.

PR #403, opened 2022-01-04
Patch 1

Whammo:
#288


jkotlinski:
I hate being nitpicking, but I see two specific problems with this change.

- `5 bank` takes 4 bytes more than `kernal-in`. I am concerned that this would waste bytes.
- The disable/enable interrupt part is missing from graphics. I suspect that will cause random crashes.

Whammo:
Not to worry, it's fun doing it.
 I'm more concerned with wasting your time.

Issue #194, opened 2017-05-05
optimize * /


jkotlinski:
They are pretty slow compared to Blazin' Forth. Reference: https://theultimatebenchmark.org/

burnsauce:
Are you talking about the two words `*` and `/` or are you talking about the word `*/`?

burnsauce:
Oof, had a look at `*`.  While durexforth's `UM*` is faster than blazin' forth's, the `*` is like 60% slower on durexforth.  Blazin's `*` is machine code, whereas durexforth uses forth.  On blazin', `*` -> `UM*` has very low overhead.

Are there differences in the '83 spec that could account for more of this shortfall?  Precision, etc.

jkotlinski:
so multiply is fixed. division is still outstanding problem, right?

PR #404, opened 2022-01-05
Optimized signed multiplication words, re: Issue #194

burnsauce:
Speed is within spitting distance of blazin' forth but adds some bytes to the core. `M*` with the same treatment might save them back.


math.asm
```diff
@@ -163,3 +163,25 @@ end:    INX
     sta MSB+1,x
     inx
     rts
+
+    +BACKLINK "abs", 3
+ABS:
+    lda MSB,x
+    bmi +
+    rts
++   jmp NEGATE
```
jkotlinski, line 172: cool! to make it a yota faster, I guess one could move NEGATE close enough to branch to it directly?

jkotlinski: ...that would also save a few bytes, to replace the `jmp NEGATE ` calls with `b?? NEGATE`

burnsauce: Done.  Indeed saves a few bytes.

burnsauce:
Sadly, there's about a 20% speed decrease from a5dc904

The hand-rolled assembly is just that much more efficient than the forth-factored code. :(

jkotlinski:
Humm... you mean 20% speed increase, or? Isn't that a good thing?

burnsauce:
Uncommitted to this codebase, I had written a lower-level `*` in machine code (that replicated negate without the jsr. It was about 20% faster than the current commit, on-par with blazin'.

forth_src/compat.fs
```diff
@@ -53,6 +53,7 @@ begin over c@ digit? over and while
 1+ r> 1- repeat ;
 
 \ from FIG UK
+: ?negate 0< if negate then ;
```
jkotlinski, line 56: It is better to remove `?negate` IMHO. It is not a standard word.

burnsauce: It's used by `SM/REM`, just below.

jkotlinski:
Oops! Sorry.

burnsauce:
I think this is it for the multiplies.  Bytes and cycles shaved, looks almost optimal to me.

burnsauce:
Only marginally slower than blazin' forth on `*` and `M*` now, although durexforth's performance gap between `*` and `M*` is wider than blazin's.

math.asm
```diff
@@ -163,3 +164,59 @@ end:    INX
...
+    +BACKLINK "dnegate", 7
+DNEGATE
+    jsr INVERT
+    inx
+    jsr INVERT
+    dex
+    lda #1
+    ldy #0
+    jsr pushya
+    jmp M_PLUS
```
burnsauce, line 217: I don't like this approach to `D1+`.  I'm working on the division right now and it's preventing some refactoring.

If you're okay with it, I'll continue division in this pull request?

burnsauce: Changed in this pull request anyway.

math.asm
```diff
@@ -163,3 +164,64 @@ end:    INX
...
+    jsr ABS
+    dex
+    jsr U_M_STAR
+    pla
+    rts
+
+    +BACKLINK "*", 1
+    jsr DABS_STAR
+    inx
+    and #$ff
```
jkotlinski, line 204: it would be a net saving to replace `and #$ff` with the `pla` from DABS_STAR. (`m*` would need to update, too)

math.asm
```diff
@@ -163,3 +164,64 @@ end:    INX
...
+    jsr INVERT
+    jmp ONEPLUS
+
+    +BACKLINK "abs", 3
+ABS
+    lda MSB,x
+    bmi NEGATE
+    rts
+
+DABS_STAR
```
jkotlinski, line 189: it would be a nice touch if input+output was documented in a comment.

math.asm
jkotlinski: If that is done, `jsr U_M_STAR , rts` can also be collapsed to `jmp U_M_STAR`.

burnsauce: This was the original design, but causes a CPU JAM when compiling `labels`.  I didn't look into it too deeply.  There's only 2 multiplies in there so ???

jkotlinski: ah, of course it doesn't work, because the return adress needs to stay on top of return stack.

jkotlinski:
good to go?

burnsauce:
Yep, all done!

jkotlinski (approved):

PR #408, opened 2022-01-06
 Separate out rs232 support to Minimize interpreter.asm

Whammo:
It's all pretty much documented. rs232 from the module works as much as it ever did.


Whammo:
I was a little over confident- hold on

PR #409, opened 2022-01-06
Update interpreter.asm

Whammo:
Not certain how this was seem-idly working before, but this takes care of the rts .
It's a testament to the stability of durexForth.


jkotlinski:
I don't think it worked before, because if I test without the fix, it crashes.

The code is however OK in v3.0.0, so this does not need to be mentioned in changelog.

PR #410, opened 2022-01-06
Faster divide

burnsauce:
Will close issues #221 and #194


burnsauce:
wait this looks wrong... ???

PR #417, opened 2022-01-07
Optimized GREATER_THAN, LESS_THAN

burnsauce:
Code reduction and overall speed improvement, although `<` is now slower than `>`, the inverse of before.


jkotlinski:
I had to revert this one, as tests were failing.

burnsauce:
Oof, signed numbers.  I could've sworn I ran `include test` :(

PR #425, opened 2022-01-09
Leveraged W's untaintedness to speed up FIND

burnsauce:
A little shortcut that saves a few cycles O(1)


burnsauce:
This is too small-potatoes to be worth this PR

PR #426, opened 2022-01-09
Factored FIND for speed

burnsauce:
It's an O(1) improvement, but it is faster.

The `find-name` implementation is big and slow, but `find` is faster.

### DIFF
--- a/forth_src/base.fs
+++ b/forth_src/base.fs
@@ -45,6 +45,8 @@ then branch ;
 begin ?dup while postpone then
 repeat ; immediate
 
+: >xt dup c@ $3f and + 1+ @ ;
+
 ( dodoes words contain:
  1. jsr dodoes
  2. two-byte code pointer. default: rts
--- a/forth_src/base.fs
+++ b/forth_src/base.fs
@@ -182,17 +182,21 @@ does> dup @ to here
 
 marker ---modules---
 
+
+.( labels..) include labels
+.( doloop..) include doloop
+.( sys..) include sys
+.( debug..) include debug
 .( wordlist..) include wordlist
 
 \ hides private words
 hide 1mi hide 2mi hide 23mi hide 3mi
 hide curr hide end hide latestxt
 hide dodoes
+hide locs hide locp hide refs hide refp
+hide lstk hide lsp hide >l hide resolve-leaves
+hide last-dump
 
-.( labels..) include labels
-.( doloop..) include doloop
-.( sys..) include sys
-.( debug..) include debug
 .( ls..) include ls
 .( require..) include require
 .( open..) include open
--- a/interpreter.asm
+++ b/interpreter.asm
@@ -363,25 +363,6 @@ FIND ; ( caddr u -- xt -1 | xt 1 | caddr 0 )
     and #STRLEN_MASK
     jmp .string_compare_failed
 
-    +BACKLINK ">xt", 3
-TO_XT
-    lda MSB, x
-    sta W + 1
-    lda LSB, x
-    sta W
-    ; W contains pointer to word
-    ldy #0
-    lda (W), y ; a contains string length + mask
-    and #STRLEN_MASK
-    clc
-    adc #1 ; offset for char + string length
-    adc LSB, x
-    sta LSB, x
-    bcc +
-    inc MSB, x
-+   jsr FETCH
-    rts
-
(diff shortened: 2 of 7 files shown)


burnsauce:
Note that this is a first-working commit at this point, I'm still testing and improving.

The general idea is that removing the well-factored forth structure from this area of code will improve performance.  Each call to FIND used to perform a bunch of repeated operations with the dictionary pointer on the forth data stack.  Several words, in turn, would take this value and index into the dictionary, forthily producing our desired XT and FLAG.

Now, W holds the dictionary pointer as the XT and FLAG are gathered.  To facilitate this, FIND holds its functionality top-to-bottom.

FIND-NAME is now in forth, and is a terrible implementation.

burnsauce:
Benchmark at this point is ~0.32 ms/FIND faster.

vice-screen-2022010917374480
vice-screen-2022010917401954

interpreter.asm
```diff
@@ -256,61 +246,38 @@ print_word_not_found_error ; ( caddr u -- )
...
 FIND_BUFFER = $33c
 FIND_BUFFER_SIZE = 31
 
-    +BACKLINK "find-name", 9
-FIND_NAME ; ( caddr u -- nt | 0 )
-    inx
-    lda LSB-1,x
+    +BACKLINK "find", 4
+    jsr COUNT
+FIND ; ( caddr u -- xt -1 | xt 1 | caddr 0 )
```
burnsauce, line 266: This is a kludge, allowing the local code to use a non-counted string.  It's also a bug according to the standard, as this code does not return the caddr but the saddr when failing.

No code in the base relies on the standard behaviour, but that could be fixed.

forth_src/base.fs
```diff
@@ -182,17 +182,21 @@ does> dup @ to here
 
 marker ---modules---
 
+
+.( labels..) include labels
+.( doloop..) include doloop
+.( sys..) include sys
+.( debug..) include debug
 .( wordlist..) include wordlist
```
burnsauce, line 190: As `find-name` is no longer in the core, it needs more facilities before loading.

interpreter.asm
```diff
@@ -375,27 +358,6 @@ FIND_NAME ; ( caddr u -- nt | 0 )
...
-    sty LSB, x ; 1
-    rts
-
-.not_immed
-    lda #$ff
-    sta LSB, x
-    sta MSB, x
-    rts
-
     +BACKLINK ">xt", 3
```
burnsauce, line 361: Used in the codebase as `latest >xt` for various self-reference.  This could possibly be factored to smaller forth code.

burnsauce (commented):
Testing this has gone well.  I do not like that FIND is not to standard, but I also don't like that it's the only thing that works with cstrings in the interpreter and it starts with COUNT.

interpreter.asm
burnsauce: Thoughts @jkotlinski ?

jkotlinski: Christmas vacations ended and so I do not have so much spare time or brain cycles anymore :-)  I still have to delve into this PR to understand what it is about.

In general, I think it is strongly preferable to be standard compliant when possible.

jkotlinski:
Sorry for not reviewing this yet! I should try to have a look at it soon.

jkotlinski:
OK, I had a quick look at it now.

One thing made me a bit puzzled, and it is, why prefer FIND over FIND-NAME? FIND is a holdover from Forth-83, it is quite odd that it takes a counted string as input.
If you look at the proposal for [FIND-NAME](https://forth-standard.org/proposals/find-name), it suggests that FIND-NAME is a "natural factors of all words that look up words in the dictionary, such as FIND, ', POSTPONE, the text interpreter, ..." I agree with the proposal and think it makes sense to have FIND-NAME as a (fast!) primitive.

jkotlinski:
Also... yes, I can see that there is some overhead with the FIND-NAME calls. But what is the performance impact in % ? 32 cycles does not sound like a lot, compared to the time it takes to search the full word list.

burnsauce:
I agree that it's a marginal improvement that breaks the standard, and am happy if this PR is rejected.

jkotlinski:
I am closing then, for reasons stated above. Thanks for the contribution anyway! Much appreciated!

PR #429, opened 2022-01-11
Adds GETIN to io.fs

Whammo:
GETIN is required for RS232 receive and is useful for input from the keyboard buffer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  - Moved and shrunk PAD area.
  - Sped up signed multiply and divide
  - Data stack effects of :/;/:NONAME/DEFINE. Starting with 2.0.0, :/:NONAME/DEFINE would put a value on the data stack, to be later consumed by ;. This is no longer the case.
+### Added
+ - GETIN in IO:
 ### Fixed
  - LOADB/SAVEB could change active device.
  - IOABORT did not print all error messages.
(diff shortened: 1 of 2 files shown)


jkotlinski:
Hey, I merged this a bit quick - now I'm having second thoughts...

First, let me start by agreeing that GETIN is useful...

What I realize now, after thinking some more, that one could add many other words that are also potential useful... like, all the other Kernal I/O functions.

I find it difficult to draw the line, what should be included and not.
For consistency, maybe it is best only to include the I/O functions that are strictly needed to build durexForth?

The rest can easily by added by user. for example. `: getin $ffe4 sys ar @ ;`

jkotlinski:
It could also be that I'm overthinking things, and it is fine just to add it. In that case, I think it should also be added to the manual.

Whammo:
Although: CHRIN with RS-232 loops until a nonnull character is found. *scratches head*

Whammo:
No, it will sit there with a full input buffer if it doesn't get a CR before that happens. It's the perfect thing for compile from rsr232, with a slight modification to .getLineFromConsole

jkotlinski:
Hey!
I'm really sorry, but after much consideration I decided to revert this merge.
The reason is that cartridge ROM space is really precious, and nothing else in Durexforth depends on this new functionality.
So I want to reserve the free space, in case something essential pops up in the future.
I think it would be great to have functions like this in an optional module, though.
Apologies again, I really want to say, I'm grateful for the contribution!

Whammo:
Not to worry- It was no work at all, simply duplicating and changing the address from CHRIN.

Issue #433, opened 2022-01-14
Idea: Double Literals and M*/


burnsauce:
Some forths feature double literals via a trailing period, e.g.: `40000.` , `-1234567.`

While double literals themselves are not that big a feature, they require implementation of `m*/` to operate.  From the standard:

> M*/ was once described by Chuck Moore as the most useful arithmetic operator in Forth. It is the main workhorse in most computations involving double-cell numbers.

It's the only operator that multiplies with a double as an input.

Should these be added to Durexforth?

jkotlinski:
What is the use case?
Personally I never had much of a desire to use double numbers.

burnsauce:
> Fixed point math would benefit greatly from M*/

The single-precision divisor limits fractional components to 14-bits, as, say fractional multiply requires `f1 * f2 / 1 << fbits`.  As we cannot divide by a signed $8000 we must set the fractional component at 14 to divide by $4000.

Therefore, `M*/` offers no great automatic advantage in fractional math for multiplication.  I will continue to explore this space.

PR #434, opened 2022-01-14
Double literals, M*/

burnsauce:
This patch does 2 things:
1. Implement `M*/`
2. Implement double literals with a trailing period

Discussion of this feature at #433 

`M*/` is required to achieve the double literal because base conversion must use doubles as input to multiplication, and this is the only forth operator to do this.  Internally it's a 32*32=64 bit multiplier with a 48/16=32 bit divider.

There is no performance impact to the math.asm changes: `ud/mod`'s changes come out in the wash when integrated with `ut/mod`, an unlisted function that accepts a triple (48-bit) dividend and a 16-bit divisor.  It is effectively the exact same routine as before, as the cascading divison required a leading 0 to divide the top word anyways.

`READ_NUMBER` is slower, owing to the 32-bit multiplication, but now able to return a double.

The code is obviously larger.

This commit updates the changelog and has cleared `include test`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  - Data stack effects of :/;/:NONAME/DEFINE. Starting with 2.0.0, :/:NONAME/DEFINE would put a value on the data stack, to be later consumed by ;. This is no longer the case.
 ### Added
  - GETIN in IO:
+ - M*/
+ - Literal doubles with a trailing period, e.g.: -100000.
 ### Fixed
  - LOADB/SAVEB could change active device.
  - IOABORT did not print all error messages.
--- a/interpreter.asm
+++ b/interpreter.asm
@@ -663,10 +663,12 @@ READ_NUMBER
 
     ldy #0
     sty .negate
+    sty .is_double
     dex
     dex
     sty LSB+1,x
     sty MSB+1,x
+    sty LSB,x
     sty MSB,x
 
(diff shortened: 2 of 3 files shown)


jkotlinski:
Hi!
Just wanted to say, sorry I haven't had the time to review yet!
It's been a busy start of the year, and I haven't had much spare time or brain cycles.
I will get to it when I can give it the attention it deserves.

burnsauce:
> Hi! Just wanted to say, sorry I haven't had the time to review yet! It's been a busy start of the year, and I haven't had much spare time or brain cycles. I will get to it when I can give it the attention it deserves.

No problem.  I update the related issue with my new doubt about use of doubles for fixed point maths.  The single-precision divisor of `M*/` seems to be a severely limiting factor.

As fixed-point was going to be my primary use of doubles, I now doubt the utility of this patch.

I'll keep thinking about it and update this patch to my latest, but there's no imperative to merge this.

jkotlinski:
OK, I will keep it open for further pondering.

jkotlinski:
Did you think further about this topic? I have been undecided, if I think this functionality would be a valuable addition or not.

To get some bearings, I did a quick investigation on how other C64 Forths do. It seems double literals are near-universally supported, while M*/ is less common, but implemented by some high-quality Forths (e.g. Blazin` Forth).

Personally I do not care so much about either, as I prefer to do any serious number crunching on the PC, and then lift the results to C64.
But I think it would be fair to argue that double literals is something one could expect from a 16-bit Forth.
As such, I am open for this idea, if anyone wants it added.

PR #436, opened 2022-04-21
Optimize EXIT, C!, C@ and 0=

KarolS:
Hi!

I found few places where I could squeeze some more bytes or cycles.

EXIT:
2 bytes, 1 cycle saved.

C!:

3 bytes, 3 cycles saved.

C@:

3 bytes, 3 cycles saved.

0=
For numbers 1-$ff, 6 cycles saved.
For numbers $100, $200,... $ff00, 6 cycles more.
Since the former are much more common than the latter, it's statistically a net saving.
Other numbers unaffected.

### DIFF
--- a/control.asm
+++ b/control.asm
@@ -113,13 +113,13 @@ EXIT
     lda HERE_LSB
     sec
     sbc #3
-    sta .instr_ptr
+    tay
     lda HERE_MSB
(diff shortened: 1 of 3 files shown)


jkotlinski:
Looks great! Thank you for the contribution!

PR #446, opened 2022-07-24
Added POSTPONE error handling. Drive-by code simplifications.

jkotlinski:


forth_src/base.fs
```diff
@@ -1,15 +1,16 @@
 : 2+ 1+ 1+ ;
 : nip swap drop ;
 : jmp, 4c c, ;
-: ['] ' [ ' literal compile, ]
+: postpone bl word dup find ?dup 0= if
+count rvs type '?' emit cr abort then
+rot drop -1 = if [ ' literal compile,
+' compile, literal ] then compile,
```
jkotlinski, line 7: At this point, I don't really understand what the `' literal compile, ` compile, literal` twister does. But it seems to work fine.

lonetech: It's still the same two logical parts. The first three words were already there, and the last three replace `['] compile,`; the brackets perform things at compile time of `postpone`, it still looks up `compile,`, and `literal` transfers the result from the compile time stack to the run time stack. The `'` and `literal` do the same job as inside the compiled `[']`. I do agree it's a little boggling to see the long string of 3 words 7 times in rapid succession. It's because we don't yet have `postpone` and `[']`; the block would read `if postpone literal ['] compile, then` with them.

lonetech:
Looks like a move of `rot` from base.fs to core.asm was mixed in. That code change also looks okay, though it wasn't mentioned. I suppose it will build faster, but there's no asm calling it, no label to do so, and base.fs still depends on asm words. The only difference at run time will be the order of words.

jkotlinski:
The move of `rot` was so it could be used in `postpone`!

forth_src/base.fs
jkotlinski: I think I get it now.

lonetech:
Thank you for the `rot` clarification, and pardon my confusion. I didn't spot that dependency. And asm in turn depends on `does>` etc so I guess this was the smaller move. Not a problem it is done, I just didn't see why at the time.

PR #452, opened 2022-07-29
Batch up c1541 commands to build disk image in one step

lonetech:
Proposed improvement for #451

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,20 +47,21 @@ durexforth.prg: *.asm
 
 $(DISK_IMAGE): durexforth.prg Makefile ext/petcom $(SRCS)
 	touch $(EMPTY_FILE)
-	$(C1541) -format "durexforth,DF" $(DISK_SUF) $@ # > /dev/null
-	$(C1541) -attach $@ -write durexforth.prg durexforth # > /dev/null
-	$(C1541) -attach $@ -write $(EMPTY_FILE) $(SEPARATOR_NAME1) # > /dev/null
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -47,20 +47,21 @@ durexforth.prg: *.asm
...
 	echo -n "aa" > build/header
 	@for forth in $(SRC_NAMES); do\
         cat build/header $(SRC_DIR)/$$forth.fs | ext/petcom - > build/$$forth.pet; \
-        $(C1541) -attach $@ -write build/$$forth.pet $$forth; \
+        echo >>c1541.script write build/$$forth.pet $$forth; \
     done;
-	$(C1541) -attach $@ -write $(EMPTY_FILE) $(SEPARATOR_NAME3) # > /dev/null
+	echo >>c1541.script write $(EMPTY_FILE) $(SEPARATOR_NAME3)
+	c1541 <c1541.script
 	rm -f $(EMPTY_FILE)
```
jkotlinski, line 65: c1541.script could be deleted here, too!

jkotlinski:
wow, brilliant idea!

Makefile
lonetech: It could. Or it could be left to make itself, by setting it up as an [intermediate](https://www.gnu.org/software/make/manual/html_node/Special-Targets.html#index-_002eINTERMEDIATE) file. It looks like that's what you're using _empty.txt for too.

jkotlinski (approved):

PR #455, opened 2022-08-02
print warning when replacing a word

jkotlinski:


jkotlinski:
I will merge this if anyone thinks it is useful :-)

Issue #448, opened 2022-07-25
Adding word index to documentation


lonetech:
Suggested labels: documentation enhancement

I wanted an index of words while browsing the documentation, and in discussion #328 there was suggestion of making the documentation accessible on the target computer as well. I've done some experiments with TeX makeindex, and while it could neatly generate an index it gets rather tricky to do all the necessary quoting for symbols together with proper sorting. This is already an issue with names like `---modules---`. Trying to define shorthand macros just exacerbated the issue, which admittedly may be because of my inexperience with TeX. 

FWIW, these are the basic commands I applied to do TeX indexing:

    \usepackage{imakeidx}
    \index{-{-}-modules-{-}-}\texttt{---modules---}
    \makeindex[intoc,columns=3]
    \printindex

The rough cut of that experiment is at https://github.com/lonetech/durexforth/tree/doc-index-tex

I ran into things like words starting with `-` being offset to the left, `@` disappearing and `2@` showing as `2`, `"` breaking the index, etc. (Both those can be escaped with a `"` in front.)

However, this is a challenge others have already faced. 

Texinfo was always designed for online documentation combined with TeX typesetting for print, specifically for programs; it has the concept of code in indices, with only three special characters `@{}` all escaped with `@`. I did a rough bit of conversion (starting point with pandoc --to texinfo) as a proof of concept, but want to open discussion before committing to converting the whole manual. pandoc makes a mess of some things, like putting # in anchors, but what's been automatically mangled can be automatically unmangled with a helpful macro in vim. I'll endeavour to document those, this is just proof of concept. 

There are clear differences I've yet to address, like some red highlights in the word anatomy section. It's definitely possible to do those, but I haven't figured out how yet. Also, words defined with the `@deffn` command have to have a category, for which I simply reused the subsection. It shows up at the right. Particular differences show at `case` (four words in one description), `pick` (uses subscripts which looked so terrible in text output I made them PDF-only), `value` where the fetch isn't actually a word definition (looks okay in PDF, but bad in html or info - need to select a solution), and under Variables where I abused the C-like return type word to keep the metavariable *foo* in the definition. 

https://github.com/lonetech/durexforth/tree/doc-index-texinfo

The texinfo version builds easily with `texi2any --pdf manual.texi`, no faffing about with multiple passes to get the index correct. 

[durexforth.pdf](https://github.com/jkotlinski/durexforth/files/9179167/durexforth.pdf)
[...]

PR #459, opened 2022-08-04
Forth word index in latex documentation

lonetech:
Builds (somewhat clumsily) HTML, ePub and PDF versions of the documentation, now with an index. 
Intended as a starting point for #448 

Multiple passes are sadly required to get ToC and index right.


docs/manual.tex
```diff
@@ -1,16 +1,34 @@
...
 \usepackage{hyperref}
 \usepackage{color}
 \usepackage{graphicx}
 \usepackage{eso-pic}
 \usepackage{multicol}
 \usepackage{alltt}
 
+\makeindex[intoc, columns=3]
+
 \hypersetup{pdfauthor="Ravelli/Durex", pdftitle="durexForth: Operators Manual", colorlinks=true}
```
jkotlinski, line 13: ("ravelli" and "durex" are joke names, at some point it should be changed for something appropriate)

docs/manual.tex
```diff
@@ -1,16 +1,34 @@
 \documentclass[a4paper]{report}
 \usepackage{microtype}
+\usepackage{imakeidx}
```
jkotlinski, line 3: How comes `imakeidx` instead of regular `makeidx`?

jkotlinski (commented):
I'm not sure I can provide much meaningful feedback. Is it fine to merge?

docs/manual.tex
lonetech: The main known difference is that imakeidx supports multiple indices (e.g. concepts and words). The reason I used it is that imakeidx supports options, in this case to be included in the table of contents and render in three columns. There exist more variants, like index, indextools, robustindex, etc. Basically, I picked the easy path. Alternate rendering tools like hevea and plastex had limited support for other packages than makeidx and imakeidx, but that's probably a non-issue in tex4ht.

docs/manual.tex
lonetech: The quotation marks here also make it into the PDF metadata, where they probably shouldn't be.

lonetech:
I think it should be fine to merge. If we find a better way to do things like the epub index, we'll put that in then. For now, there's more to do and I'd love to have this as a basis.

jkotlinski:
Cool! Thank you for the contribution!

jkotlinski:
@lonetech after getting the changes, I get this error when building the manual. What am I missing?

`Writing index file durexforth.idx
(/usr/share/texmf-dist/tex/latex/l3backend/l3backend-pdftex.def)
(./durexforth.aux) (/usr/share/texmf-dist/tex/latex/microtype/mt-cmr.cfg)
(/usr/share/texmf-dist/tex/latex/hyperref/nameref.sty
(/usr/share/texmf-dist/tex/latex/refcount/refcount.sty)
(/usr/share/texmf-dist/tex/generic/gettitlestring/gettitlestring.sty))
(./durexforth.out) (./durexforth.out)
(/usr/share/texmf-dist/tex/context/base/mkii/supp-pdf.mkii
[Loading MPS to PDF converter (version 2006.09.02).]
) (/usr/share/texmf-dist/tex/latex/epstopdf-pkg/epstopdf-base.sty
(/usr/share/texmf-dist/tex/latex/latexconfig/epstopdf-sys.cfg))
! Undefined control sequence.
l.18 \ifdef
           {\Css}{
?
`

lonetech:
Oh my. That's not what we want; the whole point of `\ifdef` is to avoid running undefined control sequences. This could be a TeX distribution difference; I'm currently using texlive 2022.20220722-1 from Debian. [texfaq.org](https://texfaq.org/FAQ-isdef) suggests `\ifdefined`. 

Could you try if https://github.com/jkotlinski/durexforth/commit/573a3597ec8a335aa102ccfa74df2750dbabc221 helps?

jkotlinski:
`\ifdefined` did not help. `\import{etoolbox}` did the trick.

jkotlinski:
@lonetech another problem now. Any idea what might be wrong?

(this uses the script at share/texmf-dist/scripts/make4ht/make4ht)

```
make4ht -e manual.mk4 -j durexforth manual.tex
[STATUS]  make4ht: Conversion started
[STATUS]  make4ht: Input file: manual.tex
This is pdfTeX, Version 3.141592653-2.6-1.40.24 (TeX Live 2022/Built by MSYS2 project) (preloaded format=latex)
 restricted \write18 enabled.
entering extended mode
...64/share/texmf-dist/scripts/make4ht/make4ht-indexing.lua:123: attempt to index a nil value (local 'f')
make: *** [Makefile:25: durexforth.html] Error 1
```

jkotlinski:
Probably I should simply install full texlive distribution. I will try that.

PR #462, opened 2022-08-05
v: fix visual bug when saving. closes #457

### LINKED ISSUE #457: Space after filename or before drive status upon save in V slipped
saving boo00, ok,00,00

save-v

jkotlinski:


Whammo:
I understand the need to suppress Kernal i/o error messages $9d, bit 7.
We have the more useful BASIC messages.

 I'm not certain it is necessary to explicitly disable Kernal i/o control messages, $9d bit 6. at all.
Only Kernal load, verify and save, it seems, have the control messages, and it doesn't interfere with ```ls```

jkotlinski:
Right now, SAVEB is the only word that temporarily modifies $9d. Overall, I think it makes sense to not touch it & leave it as an user option.

PR #480, opened 2022-09-08
Minor edits of AsciiDoc manual

lonetech:
Mostly formatting tweaks. The assembly mnemonics page is probably debatable taste. The book doctype adds a title page, which might be desirable to adjust further (e.g. it supports background image).


jkotlinski:
Thank you for the contribution :)

jkotlinski:
Had a look at it now, real neat! I love the kbd: macro!

Right, the assembly mnemonics layout is not easy.

Maybe it would be a good idea to switch to the traditional Ragsdale assembler, where multi-mode opcodes such as LDA are augmented by address mode words such as # ,X ,Y ) .A ...
It would reduce word count, and make it easier to format the documentation nicely.

PR #506, opened 2023-01-15
 Superfluous clrchn in SEND-CMD

Whammo:
Default channels already set, Typing and seeing it in on screen. before calling ```send-cmd```


jkotlinski:
approved!

PR #508, opened 2023-01-15
CLRCHN after close_all_logical_files in interpreter.asm

Whammo:
close_all_logical_files should call CLRCHN .

### DIFF
--- a/asm/disk.asm
+++ b/asm/disk.asm
@@ -273,15 +273,3 @@ INCLUDED
     jmp -
 +   jmp CLOSE_INPUT_SOURCE
 
-; Used registers: A, X, Y
-close_all_logical_files:
-    jsr CLRCHN
-    ldx #0
--   txa
-    pha
-    jsr CLOSE
-    pla
-    tax
-    dex
-    bne -
-    rts
--- a/asm/disk.asm
+++ b/asm/disk.asm
@@ -275,6 +275,7 @@ INCLUDED
 
 ; Used registers: A, X, Y
 close_all_logical_files:
+    jsr CLRCHN
     ldx #0
 -   txa
     pha
--- a/asm/disk.asm
+++ b/asm/disk.asm
@@ -111,7 +111,7 @@ load_binary_laddr_hi = *+1
     lda #0		;0 = load to memory (no verify)
     jsr LOAD
     bcs .disk_io_error
-    jmp CLRCHN
+    rts
 
 .disk_io_setnamsetlfs ;reused by both loadb and saveb
     jsr SETNAM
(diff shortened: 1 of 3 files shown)


Whammo:
I can't find more instances of jsr close_all_logical_files
  Did I miss one?

jkotlinski:
I think this makes a lot of sense! close_all_logical_files tries to emulate C128 CLALL, which indeed calls CLRCHN.

asm/interpreter.asm
```diff
@@ -79,8 +79,7 @@ quit_reset
     stx     SAVE_INPUT_STACK_DEPTH
 
     jsr     close_all_logical_files
```
jkotlinski, line 81: It looks like close_all_logical_files is only used here. So it could even make sense to inline it.

asm/disk.asm
```diff
@@ -111,7 +111,7 @@ load_binary_laddr_hi = *+1
     lda #0		;0 = load to memory (no verify)
     jsr LOAD
     bcs .disk_io_error
-    jmp CLRCHN
+    rts
```
jkotlinski, line 114: I think the intent of this CLRCHN call was to switch back to use keyboard as input device. Is it wrong?

Whammo: LOADB is called by, LS and V. Both are using keyboard and screen.
  LOAD doesn't change the active devices.

asm/interpreter.asm
```diff
@@ -78,9 +78,18 @@ quit_reset
...
-    jsr     close_all_logical_files
-    jsr     CLRCHN
-
+; Used registers: A, X, Y
+close_all_logical_files:
+    jsr CLRCHN
+    ldx #0
+-   txa
+    pha
+    jsr CLOSE
```
jkotlinski, line 87: is it safe to call CLRCHN before CLOSE? Usually, CLOSE is called before CLRCHN?

jkotlinski: Maybe the order does not matter? I do not remember anymore how this works :-)

Whammo: You should call CLRCHN before CLOSE. Especially if the device you want to CLOSE is a current device.
 Otherwise, one of your current devices will be invalid after close.

asm/disk.asm
Whammo: load_binary may be inlined too.

asm/interpreter.asm
Whammo: It is also true that these fix no problems, and the memory savings are likely inconsequential.

jkotlinski:
approved!

PR #517, opened 2023-02-04
Just a few comments for new readers of the source

ekipan:
Take 2, as per https://github.com/jkotlinski/durexforth/pull/514#issuecomment-1416812515.

### DIFF
--- a/asm/durexforth.asm
+++ b/asm/durexforth.asm
@@ -4,10 +4,7 @@
 
 !cpu 6510
 !to "durexforth.prg", cbm	; set output file and format
-
-* = $801
-
-!byte $b, $08, $a, 0, $9E, $32, $30, $36, $31, 0, 0, 0 ; basic header
+!ct pet
 
 ;; Word flags
 F_IMMEDIATE = $80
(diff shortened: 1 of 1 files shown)


asm/durexforth.asm
```diff
@@ -26,38 +40,38 @@ TIB = $200
...
+!set __LATEST = WORDLIST_BASE
+!macro BACKLINK .name , .namesize {
+    !set .xt = *
+    * = __LATEST - len(.name) - 3
+    !set __LATEST = *
+    !byte .namesize
+    !text .name
+    !word .xt
+    * = .xt
+}
```
ekipan, line 73: While moving this macro I also changed the definition. I hope it's clearer. Not that it wasn't really before.

jkotlinski:
Thank you!

jkotlinski:
By the way. What did you think of the idea to move durexforth.asm to root folder, to make it easier for new users to find the entry point?

ekipan:
I'm torn. On the one hand, yeah, it makes it more visible but on the other it's nice to have the sources together. Don't know if I would.

PR #520, opened 2023-02-06
added `compare`

ekipan:
I believe it to be ANS Forth correct. I haven't looked at your testing rig but it'd be nice to use the official tests.

This is probably much better done in direct assembly though.

### DIFF
--- a/forth/compare.fs
+++ b/forth/compare.fs
@@ -12,15 +12,3 @@ dup 0 > if drop 1 exit then
 dup 0< if drop -1 then ;
 
 : compare strcmp signum ;
-
-(
-: !! 0= abort" fail" '.' emit ;
-: abc s" abc" drop ;
-: abd s" abd" drop ;
-abc 0 abd 0 compare 0= !!
-abc 3 abc 3 compare 0= !!
-abc 2 abd 2 compare 0= !!
-abc 3 abd 2 compare 0 > !!
-abc 3 abd 3 compare 0< !!
-.s (
-)
--- a/forth/compare.fs
+++ b/forth/compare.fs
@@ -7,8 +7,6 @@ swap 1+ swap 1+ loop  2drop 0 ;
 rot swap 2dup - >r min memcmp r>
 over if drop else nip then ;
 
-: signum ( n1 -- -1/0/1 )
-dup 0 > if drop 1 exit then
-dup 0< if drop -1 then ;
-
-: compare strcmp signum ;
+: compare strcmp
+dup 0 > if drop 1 exit then 0< ;
+( todo extract signum word? )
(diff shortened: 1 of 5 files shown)


ekipan:
Bleh, The tests worked in VICE when I tried them from v directly. Cycles again. I do not know what could be wrong.

jkotlinski:
The build archived a screenshots from the test run: 
bild

ekipan:
Probably needs to be in the Makefile? Gimme a sec.

ekipan:
Whoops, maybe I should actually read this thing before committing.

jkotlinski:
The word should probably be described in manual, in "Strings" section.

forth/compare.fs
```diff
@@ -0,0 +1,14 @@
...
+if  >r 2drop r> unloop exit  then
+swap 1+ swap 1+ loop  2drop 0 ;
+
+: strcmp ( a1 n1 a2 n2 -- diff )
+rot swap 2dup - >r min memcmp r>
+over if drop else nip then ;
+
+: signum ( n1 -- -1/0/1 )
+dup 0 > if drop 1 exit then
+dup 0< if drop -1 then ;
```
jkotlinski, line 12: `dup 0< if drop -1 then` can probably be replaced by `0<`

ekipan:
Clearly the PR is half-baked. Would you prefer I retract for now?

jkotlinski:
It is fine to keep it open. You can convert it to Draft if you like.

ekipan:
(The force push was an extra blank line in tests.fs I deleted)

ekipan:
Would you prefer file `compare` to be an optional include, as I have here, or try putting it into `base`? I heard the cartridge was tight on bytes.

ekipan:
I wrote this word and thought "hey that seems useful" and submitted on a lark, but really memcmp should be assembler, and I'm sure it already exists in the interpreter somewhere so it would be better to reorganize and expose that, but that's more work than I'm able to do at my level. Thanks for humoring me though.

Whammo:
Would you could Open a Discussion, maybe in Show-and-tell, show your code and document it there? A straight memory compare is a useful thing and offers many opportunities for use-specific optimization. I am certain I would be grateful, and there might even be a half a dozen other people in the world interested.

jkotlinski:
Hey! Sorry for not responding promptly. I had too much real-life things going on lately.

It is true, reuse value would be higher if the word was written efficiently in assembly. I would look at cc65 strcmp/memcmp for inspiration.

I do not think durexForth already has string matching code that is suitable for reuse. It's either too optimized (interpreter.asm) or not particularly good or fast (v.fs).

PR #524, opened 2023-02-14
Suggest changes to manual words.adoc RE. iec.fs

Whammo:
I thought I'd see if this was something we want to roll with before I do any more work.
What do you think?

### DIFF
--- a/manual/words.adoc
+++ b/manual/words.adoc
@@ -337,6 +337,39 @@ Low-level device I/O may interfere with disk accesses done by durexForth and the
 - If multiple files are open, always call `clrchn` to end any serial bus transactions before calling `open` or switching between files with `chkin` or `chkout`.
 ====
 
+==== Protocol-Level IEC Device I/O
+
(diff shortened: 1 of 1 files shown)


jkotlinski:
I cannot judge its merit because I do not understand this low-level stuff yet. But it looks cool!

About "These commands should never be used by anyone. Ever." why should it not be used?

Whammo:
It's half joking and half Reverse psychology. 
If it is forbidden, then perhaps it might be more desired.  😉

jkotlinski:
:-) it would maybe be more helpful to mention some typical use cases for
these words.

Or add a pointer to
http://www.codebase64.org/doku.php?id=base:io_programming for examples of
things that can be done

tis 14 feb. 2023 kl. 22:22 skrev Whammo ***@***.***>:

Whammo:
I've made a few additions and changes. If anyone would enjoy taking a look at the material and letting me know the obstacles or gaps, up or within: I'd be happy to break them down or build them up, respectively.
  All of the examples are working.

Issue #525, opened 2023-02-14
save-input/restore-input


popeyeotaku:
Could we make save-input/restore-input regular words the user could access? Maybe bump their stacks up a couple entries. They are in the standard, and you do have them implemented.

jkotlinski:
If you have a use for these words, I think it should be fine to put them in dictionary. The only reason not to have them in dictionary is to save RAM.

Ideally, SAVE-INPUT should ABORT on stack overflow.

I am not sure if it is standard-compliant to save the state in a separate stack. But I think it should be.

jkotlinski:
Actually, I read up a bit and I think some rework is needed to make these words standard compliant:

- The SAVE_INPUT_STACK should be removed.
- SAVE_INPUT/RESTORE_INPUT should put the state on the data stack, as described in the Forth standard.
- Each call to SAVE_INPUT should be followed by a call to N>R , to avoid stack overflow.
- Similarly, each call to RESTORE_INPUT should be preceded by a call to R>N.

jkotlinski:
For fun, implemented `N>R` `NR>` which is part of the solution.

The real tricky part is moving around code so that SAVE_INPUT/RESTORE_INPUT calls are in the same word. (this is needed to allow `N>R` `NR>` store data on the return stack)

jkotlinski:
OK... I read up a bit further. This is what Forth standard says: "`SAVE-INPUT` and `RESTORE-INPUT` are intended for repositioning within a single input source"

So, it looks like durexForth `SAVE-INPUT`/`RESTORE-INPUT` have a completely different functionality than intended in the standard.

They should be renamed to something else, to avoid confusion.

jkotlinski:
I now renamed SAVE/RESTORE_INPUT to PUSH/POP_INPUT_SOURCE, to avoid confusion with the standard words.

PR #526, opened 2023-02-15
Draft: standard SAVE-INPUT, RESTORE-INPUT

jkotlinski:
see: #525


jkotlinski:
abandoned. standard SAVE-INPUT, RESTORE-INPUT is not a good idea.

PR #531, opened 2023-03-14
optimize `hold`

ekipan:
Instead of moving characters forward, store them going backward.

### DIFF
--- a/forth/format.fs
+++ b/forth/format.fs
@@ -1,11 +1,9 @@
-variable end
-create #>buf #34 allot
-: <# #>buf end ! ;
-: #> 2drop #>buf end @ over - ;
-: hold
-\ reserve space for char at start
-#>buf dup 1+ end @ #>buf - move
-1 end +! #>buf c! ;
+variable holdp
+#34 allot here dup
+: <# literal holdp ! ;
+: #> 2drop holdp @ literal over - ;
+: hold -1 holdp +! holdp @ c! ;
+
 : sign 0< if '-' hold then ;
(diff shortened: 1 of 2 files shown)


forth/format.fs
```diff
@@ -1,11 +1,9 @@
...
-: #> 2drop #>buf end @ over - ;
-: hold
-\ reserve space for char at start
-#>buf dup 1+ end @ #>buf - move
-1 end +! #>buf c! ;
+variable holdp
+#34 allot here dup
+: <# literal holdp ! ;
+: #> 2drop holdp @ literal over - ;
+: hold -1 holdp +! holdp @ c! ;
```
jkotlinski, line 5: Isn't there a risk for buffer overrun with this solution? When calling `HOLD` more than 34 times, `holdp` would move past beginning of buffer?

jkotlinski: ...maybe that is fine, though? I think I am in favor of this solution.

ekipan: Same problem was in the original as well. After 34 holds it would start to overwrite the code for `<#`.

ekipan:
Though I guess the failure mode is quite different. 35th hold used to just trash `<#`, now it bops `holdp`'s MSB I think? So subsequent holds are probably a bit more destructive.

ekipan:
You could make it less unpredictable by moving `holdp` after the buffer, if you care enough.

jkotlinski:
It does not matter so much to me, if it fails one way or another :-)  I think it is fine the way it is. Thank you for the contribution!

PR #546, opened 2023-04-19
added SPLIT, removed 100/

jkotlinski:
Goal: Remove 100/, which is badly named.

- 100 no longer makes sense, because durexForth now uses DECIMAL base by default
- / does not make total sense, because it implies arithmetic right shift.

Discussion in https://github.com/jkotlinski/durexforth/discussions/543

### DIFF
--- a/forth/base.fs
+++ b/forth/base.fs
@@ -34,7 +34,7 @@ immediate
 r> 1+ count 2dup + 1- >r ;
 
 ( "0 to foo" sets value foo to 0 )
-: (to) >r split r@ c! r> 2+ c! ;
+: (to) >r split r@ 2+ c! r> c! ;
 : to ' 1+ state c@ if
 postpone literal postpone (to) exit
 then (to) ; immediate
@@ -78,12 +78,9 @@ parse-name asm included
    0 value foo
    foo . \ prints 0
    1 to foo
-   foo . \ prints 1
-
-  ! other words, such as TO, write
-  directly to the ldy/lda arguments. )
+   foo . \ prints 1 )
 : value ( n -- )
-code split ldy,# lda,#
+code split swap lda,# ldy,#
 ['] pushya jmp, ;
 : constant value ;
 ( to free up space, pad could be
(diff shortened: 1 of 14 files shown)


forth/base.fs
```diff
@@ -80,7 +80,7 @@ parse-name asm included
    1 to foo
    foo . \ prints 1 )
 : value ( n -- )
-dup code lda,# 100/ ldy,#
+code split swap lda,# ldy,#
```
ekipan, line 83: How about `split ldy,# lda,#`?

jkotlinski:
I tried that first, but I couldn’t get it to work, very tricky! Other words
assume that lda and ldy are in that order. At least TO does.

tors 20 apr. 2023 kl. 00:21 skrev ekipan ***@***.***>:

forth/base.fs
jkotlinski: I tried that first, but I couldn’t get it to work, very tricky! Other words
assume that lda and ldy are in that order. At least TO does.

I agree it would be nicer to have it that way, to avoid the swap...

jkotlinski: ok, I think I fixed up all places that needed it :-)

ekipan: Nice! :+1:

jkotlinski: Not sure I like the change in retrospect... It does save three bytes, but for a little-endian platform, it feels ugly that the most significant byte would have the lower address.

ekipan: I tend to agree. I just saw the phrase `swap [use2] [use1]` and thought `[use1] [use2]` without a second thought for the consequences. More than the three bytes, is `value` used enough times in compiling the sources that 4 extra load/stores is even >0.1% time save? Probably not?

jkotlinski: I measured with VICE: the swap makes system compilation 3 jiffies slower. That is less than 0.1%.

Whammo:
A ```join``` is known to be useful in certain disk operations.

PR #556, opened 2023-05-05
Multi-line EVALUATE

### LINKED ISSUE #545: `\` broken when compiling from `v`
Using the [latest build](https://github.com/jkotlinski/durexforth/actions/runs/4713236155) cartridge, pasting the following program into `v` and pressing `F7` won't compile the `bar` word.
```forth
: foo ;
\ test
: bar ;
```
The standard says [`\` discards](https://forth-standard.org/standard/core/bs) the rest of the parse area, so if I understand correctly that `v` compiles with `evaluate` then `\` comments no longer work.

Maybe it's rude to piggyback but an unrelated issue: the program `( test )` leaves 2 junk values on the stack. Looks like you need a `2drop` in the [definiton of `(`](https://github.com/jkotlinski/durexforth/blob/485e245f4c56b6894ac3e8e762b1ba7efad85b5f/forth/base.fs#L28).

jkotlinski:
Closes #545.

### DIFF
--- a/asm/io.asm
+++ b/asm/io.asm
@@ -167,8 +167,10 @@ EVALUATE_STRING_PTR_MSB = * + 1
     cmp #$d
     beq .return_true
 
-    inc TIB_SIZE ; assumes max line length = 255
-
+    inc TIB_SIZE
+    bne +
+    inc TIB_SIZE + 1
++
 EVALUATE_STRING_SIZE_LSB = * + 1
     lda #0
 EVALUATE_STRING_SIZE_MSB = * + 1


ekipan:
I don't have any experience implementing this stuff and only tentatively understand much of the code, to say nothing of considering all the other implementors of the Forth standard, so I'm hesitant to offer opinions. I guess the short of it is that I'm happy v can compile my code. 

But since you asked, here they are, weakly-held and also contradictory: (1) Anton has a point that it should probably have a different name but (2) honestly I don't see much point in an EVALUATE that doesn't understand multiple lines but (3) the whole REFILL SOURCE-ID scheme seems like extra complexity to support files that might be larger than RAM so is necessary but (4) the standard EVALUATE is simpler and that's appealing but (5) I also see the value in more consistency but but but etc.

I don't think I'm qualified here. Thanks for fixing v!

asm/io.asm
```diff
@@ -137,6 +127,55 @@ REFILL ; ( -- flag )
...
+    lda EVALUATE_STRING_SIZE_LSB
+    bne +
+    dec EVALUATE_STRING_SIZE_MSB
++   dec EVALUATE_STRING_SIZE_LSB
+
+    tya
+    cmp #$d
+    beq .return_true
+
+    inc TIB_SIZE ; assumes max line length = 255
```
ekipan, line 170: ... this is an assumption not in standard EVALUATE, and it's one that I think breaks my own personal BLOCK wordset that defines `: load block 1024 evaluate ;`. Maybe it's better to have a separate word after all :sweat:

jkotlinski:
Right, there is a lot of tradeoffs and I changed my mind so many times.

What made me convinced about having a custom REFILL behavior for reading
from RAM buffers line-by-line, is that Mitch wrote that it is the proper
way to do it.

About if multiline EVALUATE is the right way forward for Forth as a
language, I guess that remains to see. Personally I think it would make a
lot of sense, to have it as an optional extension in FILE wordset.

tis 9 maj 2023 kl. 07:38 skrev ekipan ***@***.***>:

asm/io.asm
jkotlinski: fixed.

PR #557, opened 2023-05-14
space-optimize `accept`

ekipan:
See [discussion](https://github.com/jkotlinski/durexforth/discussions/538#discussioncomment-5898961).


ekipan:
whoops I just notice that the rewrite rejects keys > $9f that the old one didn't. One sec.

jkotlinski:
Seems to work!

PR #560, opened 2023-05-19
renamed ;CODE => END-CODE

### LINKED ISSUE #559: ;CODE / END-CODE confusion
the durexForth assembler differs from forth-standard.org in subtle but important ways.

 * According to [CODE](https://forth-standard.org/standard/tools/CODE), the word `END-CODE` typically ends assembly.
 * According to [;CODE](https://forth-standard.org/standard/tools/SemiCODE), it should work like `DOES>` but for assembly.
 * durexForth `;CODE` does not simply work like `END-CODE`: it also compiles an `rts` instruction.

It would probably be good to iron out some or all of these quirks.

jkotlinski:
goal: confirm with forth-standard.org. closes #559

 - standard ;CODE is like DOES>, but for assembler
 - END-CODE is mentioned as the traditional way to exit a CODE segment


jkotlinski:
not the right idea.

PR #563, opened 2023-05-21
Fix tick

ekipan:
[This commit](https://github.com/jkotlinski/durexforth/commit/9c906de02985912c115f16b57d041035a9f82dd8) broke error reporting when `'` was used with a missing word. This PR fixes that. The 0 from `find-name` remained on the stack where `notfound` interprets it as the string length.


ekipan:
Looks like that was the only asm use of NIP. Should I move it back to base.fs?

jkotlinski:
Oh! Sorry for that, it was sloppy not to test enough.

I wouldn't worry about NIP, it is not out of place in core.asm.

PR #564, opened 2023-05-30
abort with error on empty HEADER name

ekipan:
See [discussion](https://github.com/jkotlinski/durexforth/commit/1fc6551d069c79ee64a8448f87b8545829bf450f#commitcomment-115573545).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,9 +20,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  - Moved tests to a separate disk (tests.d64).
  - gfx: Renamed ERASE to PEN.
  - #> (the pictured numeric output string buffer) now uses its own buffer chopped from the end of PAD.
- - CHAR, HEADER and other parsing words no longer do REFILL on missing text.
  - EVALUATE now accepts multi-line strings.
  - ( refills to accept multi-line comments when parsing EVALUATE strings.
+ - CHAR, HEADER and other parsing words no longer do REFILL on missing text.
+ - Defining words based on HEADER instead abort with an error.
 ### Removed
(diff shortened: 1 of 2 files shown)


asm/compiler.asm
```diff
@@ -136,17 +136,20 @@ COLON
     jmp RBRAC ; enter compile mode
 
 
-; --- HEADER ( name -- )
+.empty_parse
+    jmp print_word_not_found_error
```
jkotlinski, line 140: What will be printed here? I guess a `?`? In that case maybe the `err` printout of .on_stack_underflow would be a better jump target.

CHANGELOG.md
```diff
@@ -20,9 +20,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  - Moved tests to a separate disk (tests.d64).
  - gfx: Renamed ERASE to PEN.
  - #> (the pictured numeric output string buffer) now uses its own buffer chopped from the end of PAD.
- - CHAR, HEADER and other parsing words no longer do REFILL on missing text.
  - EVALUATE now accepts multi-line strings.
  - ( refills to accept multi-line comments when parsing EVALUATE strings.
+ - CHAR, HEADER and other parsing words no longer do REFILL on missing text.
+ - Defining words based on HEADER instead abort with an error.
```
jkotlinski, line 26: These two lines should probably be joined.

ekipan: How about putting "CHAR doesn't REFILL" back where it was and just having the note about "HEADER etc ABORTs" be separate?

jkotlinski: I do not like my original wording.
How about...

 * CHAR will no longer REFILL when reaching the end of input buffer.
 * Word-creating words will now abort with an error when an unexpected end of input buffer results in an empty word name.

Issue #579, opened 2024-12-12
`abort"` space savings, also github build problems


ekipan:
`v.fs` uses `abort"` six times. Each time it compiles to:
```forth
if rvs s" ..." type cr abort then
```
I propose pulling a runtime factor, say, `aborts`:
```forth
: aborts ( au-) rvs type cr abort ;
: abort" ( -;f-) postpone if
postpone s" postpone aborts
postpone then ; immediate
```
Saving 4 bytes of code plus 9 bytes per use. Personally I think `aborts` is a more useful factor.
```forth
\ say I want to reset some state, I currently have to:
: ... ( ... ) is-bad? if reset-state 1 abort" bad" then ( ... ) ;
\ which is wasteful versus:
: ... ( ... ) is-bad? if reset-state s" bad" aborts then ( ... ) ;
````
Though introducing a new word does mean more documentation, so I wouldn't oppose `hide`ing it.

Edit: had `cr` in the wrong place.

**Comment by jkotlinski** [comment]
Sounds good to me!

tors 12 dec. 2024 kl. 23:16 skrev ekipan ***@***.***>:

**Comment by Whammo** [comment]
Ioabort has, for some time returned basic io error messages. This perhaps may have lessened the need for hard coded, disk error messages.

**Comment by Whammo** [comment]
Sorry, yeah I see they're internal messages.
 Heh. Never mind.

**Comment by ekipan** [comment]
I don't actually have a means to develop locally right now. Github does have a fancy in-browser vscode application that lets you directly develop on a github remote repo. When I synced my fork the build failed tho. Here's the output that differs on [your latest build](https://github.com/jkotlinski/durexforth/actions/runs/11922426001/job/33228648497#step:3:590) vs. [my build](https://github.com/ekipan/durexforth/actions/runs/12325292773/job/34404349370#step:3:590).

```
The following packages have unmet dependencies:
 headlessvice : Depends: libflac8 but it is not installable
```

I admit I don't really know what I'm doing. I could have done something wrong, or some remote software source might have linkrotted, breaking your next build as well?

jkotlinski:
@ekipan I rebuilt the latest build, and it works for me. I don't know what I am doing either, so I do not see why it works for me but not you.

jkotlinski:
@ekipan I bumped vice-headless dependency, does it make things any better for you?

ekipan:
[...]

PR #581, opened 2024-12-14
factor out `aborts` #579

ekipan:
Thanks for build fix.

### DIFF
--- a/forth/base.fs
+++ b/forth/base.fs
@@ -144,9 +144,9 @@ here latest >xt 1+ (to)
 : .s depth begin ?dup while
 dup pick . 1- repeat ;
 
-: aborts rvs type cr abort ;
+: (aborts") rvs type cr abort ;
 : abort" postpone if
-postpone s" postpone aborts
+postpone s" postpone (aborts")
 postpone then ; immediate
 
 ( linked list. each element contains
@@ -170,7 +170,7 @@ marker ---modules---
 \ hides private words
 hide 1mi hide 2mi hide 23mi hide 3mi
 hide holdp hide latestxt
-hide dodoes hide aborts
+hide dodoes hide (aborts")
 
 .( labels..) include labels
 .( doloop..) include doloop
--- a/forth/base.fs
+++ b/forth/base.fs
@@ -144,12 +144,9 @@ here latest >xt 1+ (to)
 : .s depth begin ?dup while
 dup pick . 1- repeat ;
 
-: abort"
-postpone if
-postpone rvs
-postpone ."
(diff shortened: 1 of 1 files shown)


ekipan:
Apparently the issue reference doesn't totally work if it's in the title so just for completeness sake:

See #579.

jkotlinski:
Nice, thanks for contributing! I will look at it in detail later, and see
how it fits in with other stuff.

lör 14 dec. 2024 kl. 09:44 skrev ekipan ***@***.***>:

jkotlinski:
OK, this seems like an improvement. I am pondering various things now.

- Is the new `ABORTS` useful somewhere in DurexForth itself? I am kind of wary about adding new words. But it would feel better if there was some immediate use for it.
- Would it make sense to code golf it harder? I.e. add `(abort")` which does all the work in one word. Check condition, and either print error message and abort, or skip error message.

ekipan:
That means every runtime-untaken `abort"` turns from a single `0branch` into rstack, counting, stack juggling etc. A bit more work. Plus I feel like you'd also need to pull a factor from `lits` to make it worth it, adding 12 cycles of `jsr rts` to every runtime string literal. Something like this?

```forth
: (lits) 1+ count 2dup + 1- ;
: lits r> (lits) >r ;

: (s") '"' parse dup c, 
here swap dup allot move ;
: s" postpone lits (s") ; immediate

\ ...

: (abort") r> (lits) >r rot 
if rvs type cr abort then 2drop ;
: abort" postpone (abort") (s") ;
immediate
```

jkotlinski:
Uhh, that was just a bad idea of mine. It is better to use `if` .. `then` for the conditional, otherwise it gets too slow.

jkotlinski:
How about merging the PR as is? But I would like one change, rename `aborts` to `(abort")`. Then it is clear that it is a helper word to `abort"`, and not a part of the official API.
That is, it does not need to be documented, and it can be later removed or changed without being a breaking change.

ekipan:
This of course does add more a bit more work in the case `abort"` _is_ given a nonzero at runtime, but considering its purpose I'm sure that's okay. 12 cycles spent in the rare case a user program fails for a 9 bytes savings every time they compile the word.

ekipan:
Or I guess only 6 cycles, since `abort` doesn't `rts` :P

jkotlinski:
Looks good to me!
Maybe at some point this will throw an exception instead.

sön 15 dec. 2024 kl. 16:04 skrev ekipan ***@***.***>:

PR #583, opened 2024-12-15
exception wordset

jkotlinski:


asm/interpreter.asm
```diff
@@ -146,26 +146,16 @@ interpret_tib
...
-    jmp PUTCHR
+.throw_stack_underflow
+    ldy #-4
+    jmp throw_y
+.throw_dictionary_overflow
+    ldy #-8
+    ; fall through
+throw_y
+    lda #$ff
+    jsr pushya
```
ekipan, line 157: Am I mistaken or are your Y's and A's mixed up? I thought `pushya` put Y in MSB and A in LSB.

jkotlinski: no wonder it didn't work! nice catch!

jkotlinski:
to solve: 

> When a [THROW](https://forth-standard.org/standard/exception/THROW) returns control to a [CATCH](https://forth-standard.org/standard/exception/CATCH), the system shall un-nest not only definitions, but also, if present, locals and input source specifications, to return the system to its proper state for continued execution past the [CATCH](https://forth-standard.org/standard/exception/CATCH).

jkotlinski:
OK, so all the tests pass, which means it basically works. But need to find a good way to keep the helpful error messages.

asm/exception.asm
```diff
@@ -0,0 +1,89 @@
...
+    jsr TO_R
+    ; restore stack
+    lda LSB,x
+    tax
+    inx
+    jsr R_TO
+    rts
+
+.print_error_and_abort
+    jsr RVS
```
ekipan, line 73: maybe put a jump vector here whose address is given as a value. The asm code could just print `err` but the forth code could later give it something proper, once `.` exists.

asm/exception.asm
```diff
@@ -0,0 +1,89 @@
...
+    inx
+    jsr TYPE
+    jmp .print_cr_and_abort
++   ; TODO print error number
+    lda #'e'
+    jsr PUTCHR
+    lda #'r'
+    jsr PUTCHR
+    jsr PUTCHR
+.print_cr_and_abort
```
ekipan, line 86: Or maybe insert a `JSR NOP` between lines 85 and 86. Later `' .` could be stored there. Just as something simple.

jkotlinski:
I messed around a little, now it prints custom error messages for some error codes.

Something still to solve: when parsing a file, and a word is not found, I want it to print the missing word.

asm/exception.asm
jkotlinski: I solved error messages in the most simple way I could imagine.

jkotlinski:
OK, I think it is more or less done now, except for all the things I did not think of yet.

I am not sure what implications this has for `ioabort`.

jkotlinski:
TODO: Update manual.

jkotlinski:
TODO: Figure out if anything ought to change related to File I/O error handling.

Whammo:
Very cool. Thank you.
Perhaps a simple example?

jkotlinski:
What do you think of this?

`: loop begin again ;
: demo ." press RESTORE" cr
['] loop catch ." caught " . ;`

When pressing RESTORE, the demo will print "caught -28", which is the exception number for "user interrupt".

jkotlinski:
Mulled things over, and it is probably O.K. if exceptions -13 (Undefined Word) and -37 (File I/O Error) print error messages before THROW, rather than in the exception handler.

jkotlinski:
Probably good to go.

jkotlinski:
TODO: print unknown exception numbers.

PR #599, opened 2025-04-25
Mega65

xlar54:
Based on the c128 port, I was able to get this working on xemu, the Mega65 emulator.  On the first run (when its building the new forth system) it will crash into the monitor and the screen will go blank.  Resetting the emulated machine and loading/running again will finish the job and go into interpret mode.   Any issues are likely in the kernal.fs / sys.fs / vcore.fs because I just copied them from the c64 files.


xlar54:
image

xlar54:
obviously v doesnt work in this build as the screen memory locations are different.

